PR #6, opened 2024-06-27
Implement Authentication using Google OAuth

j-maynard:
This adds authentication against Googles OAuth provider.  With this implemented it should be straight forward to add other authentication providers to our App such as ActiveDirectory, OneLogin For Governement or social logins such as FaceBook.

We need to properly implement user storage and user management in to the system along with RBAC.

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,6 +62,7 @@
     "dotenv": "^16.4.4",
     "ejs": "^3.1.9",
     "express": "^4.19.2",
+    "express-rate-limit": "^7.3.1",
     "express-session": "^1.18.0",
     "express-user": "^1.2.0",
     "govuk-frontend": "^5.1.0",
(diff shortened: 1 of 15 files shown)


package.json
```diff
@@ -26,11 +26,12 @@
   "devDependencies": {
     "@shopify/eslint-plugin": "^44.0.0",
     "@types/better-sqlite3": "^7.6.9",
-    "@types/express": ">=4.17.21",
+    "@types/express": "^4.17.21",
     "@types/fs-extra": "^9.0.13",
     "@types/jest": "^29.5.12",
     "@types/multer": "^1.4.11",
     "@types/node": "^20.11.27",
+    "@types/passport-google-oauth20": "^2.0.16",
```
easternbloc, line 34: I assume this is here as we only care about using google auth in dev?

easternbloc (approved):

package.json
j-maynard: Thats there because its a typescript "type" definition.  Its not needed for the node app to compile and run.  It just provides the type definitions to the typescript transpiler.

PR #11, opened 2024-09-06
Make the frontend talk to the updated backend

j-maynard:
This makes the frontend talk to the refactored backend.  There are 2 tests which aren't implemented yet while we workout how to mock and handle sessions in tests.


src/views/partials/top.ejs
```diff
@@ -68,7 +68,7 @@
                         </li>
                         <li class="govuk-header__navigation-item govuk-header__navigation-item--active">
                             <a class="govuk-header__link"
-                                href="/<%= i18n.language %>/list">
+                                href="/<%= i18n.language %>/<%= t('routes.dataset') %>">
```
wheelsandcogs, line 71: We're translating the URL path as well as the labels?

wheelsandcogs (approved):

src/views/partials/top.ejs
j-maynard: We are... its part of the welsh first thing... Remind me next week.  We did some workshops on this and theres some bits around about it which might be helpful

src/dtos2/processedcsv.ts
```diff
@@ -0,0 +1,21 @@
+import { DatasetDTO } from './dataset-dto';
+import { Error } from './error';
+
+export interface PageInfo {
+    total_records: number | undefined;
+    start_record: number | undefined;
+    end_record: number | undefined;
+}
```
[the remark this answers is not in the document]
j-maynard: I did but only after I wrote this bit of code.  I'll change when I next look at the frontend (so probably next week)

PR #12, opened 2024-09-16
Update frontend to support confirming Dataset and Labelling Sources

j-maynard:
As per the ticket... Once we have confirmed a fact table is the one the user wants we then go on to labelling the sources as Dimensions, DataValues, Footnotes or Ignored.  This joins up the journey from File upload to dimension creation.


src/middleware/translations/en.json
```diff
@@ -0,0 +1,177 @@
+{
+    "greeting": "Hello",
+    "share-page": "Share this page",
+    "top-of-page": "Back to top",
+    "app-running": "App is running",
+    "health-notes": "Expand endpoint to check for database connection and other services.",
+    "developer_warning": "<strong class=\"govuk-tag govuk-tag--red govuk-phase-banner__content__tag\">Warning</strong> This is a developer page and has not been through user research",
+    "developer_tag": "<span class=\"govuk-tag govuk-tag--yellow govuk-phase-banner__content__tag\">Developer Page</span>",
```
wheelsandcogs, line 8: When we export this to send to the translators, will we strip the HTML or include it?

src/services/stats-wales-api.ts
```diff
@@ -179,14 +174,165 @@ export class StatsWalesApi {
...
+                body: formData
+            }
+        )
+            .then((response) => {
+                if (response.ok) {
+                    return response.json();
+                }
+                const err = new HttpError(response.status);
+                err.handleMessage(response.text());
+                throw err;
```
wheelsandcogs, line 317: Lots of very similar error handling code in this class, maybe we can stick a wrapper around fetch (or use a library?) and reduce the amount of boilerplate in each function.

src/services/stats-wales-api.ts
```diff
@@ -179,14 +174,165 @@ export class StatsWalesApi {
...
+            })
+            .then((api_res) => {
+                const datasetDTO = api_res as DatasetDTO;
+                return {
+                    success: true,
+                    dataset: datasetDTO
+                } as UploadDTO;
```
wheelsandcogs, line 324: What's the reasoning behind wrapping the returned values with a success flag? Isn't the alternative an exception?

src/views/partials/header.ejs
```diff
@@ -98,11 +99,19 @@
...
\ No newline at end of file
+    </div>
+
+    <% if (locals.developerPage) { %>
+    <div class="govuk-width-container app-width-container">
+        <div class="govuk-phase-banner">
+            <p class="govuk-phase-banner__content"><%- t('developer_warning')%></p>
+        </div>
+    </div>
+    <% } %>
```
wheelsandcogs, line 117: I wouldn't even bother making this conditional, this will be around for years and we want it on every page I would imagine?

src/views/publish/sources.ejs
```diff
@@ -0,0 +1,51 @@
...
+    .source-list {
+        margin-bottom: 2em;
+    }
+
+    .source-list-item {
+        border-bottom: 1px solid #0b0c0c;
+        padding-bottom:0.5em;
+        margin-bottom: 0.5em;
+    }
+</style>
```
wheelsandcogs, line 50: What's the reason for this not being in the css file?

src/views/publish/start.ejs
```diff
@@ -17,4 +18,4 @@
 	</main>
 </div>
 
-<%- include("../partials/bottom"); %>
\ No newline at end of file
+<%- include("../partials/footer"); %>
```
wheelsandcogs, line 21: Looks like this file needs a newline adding at the end - maybe your new editor doesn't have the newline on save option turned on?

src/views/publish/title.ejs
```diff
@@ -1,38 +1,21 @@
...
-                                <% errors.forEach(function(error) { %>
-                                    <li>
-                                        <a href="#<%= error.field %>"><%= t(error.tag.name) %></a>
-                                    </li>
-                                    <% }); %>
-                            </ul>
-                        </div>
-                    </div>
-                </div>
-            <% } %>
```
wheelsandcogs, line 31: 👍

test/publish.test.ts
```diff
@@ -0,0 +1,605 @@
...
+jest.mock('../src/middleware/ensure-authenticated', () => ({
+    ensureAuthenticated: (req: Request, res: Response, next: NextFunction) => next()
+}));
+
+describe('Publisher Journey Tests', () => {
+    beforeAll(() => {
+        server.listen({
+            onUnhandledRequest: ({ headers, method, url }) => {
+                if (url.includes('http://example.com:3001')) {
+                    console.log('Request to unahndled URL:', method, url);
```
wheelsandcogs, line 36: typo

tools/copyAssets.ts
```diff
@@ -4,4 +4,5 @@ import * as shell from 'shelljs';
 shell.cp('-R', 'src/views', 'dist/');
 shell.cp('-R', 'src/public/assets', 'dist/');
 shell.cp('-R', 'src/public/css', 'dist/');
-shell.cp('-R', 'src/resources', 'dist/');
+shell.cp('-R', 'src/middleware/translations', 'dist/');
+shell.cp('-R', 'src/middleware/translations', 'dist/middleware/translations');
```
wheelsandcogs, line 8: Is this copying the translation files to both the root of the compiled app and the translation dir? Do we need it in both locations?

src/routes/publish.ts
```diff
@@ -8,196 +8,602 @@ import { StatsWalesApi } from '../services/stats-wales-api';
...
+        }
+        setCurrentToSession(viewDTO.dataset, req);
+        res.redirect(
+            `/${lang}/${req.i18n.t('routes.publish.start', { lng: lang })}/${req.i18n.t('routes.publish.preview', { lng: lang })}`
+        );
+    } else {
+        const err = processedCSV as UploadErrDTO;
+        req.session.errors = generateViewErrors(err.dataset?.id, 500, err.errors);
+        res.redirect(`/${lang}/${req.i18n.t('routes.publish.start', { lng: lang })}/`);
+    }
```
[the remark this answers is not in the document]
j-maynard: Switched this around

src/routes/publish.ts
```diff
@@ -8,196 +8,602 @@ import { StatsWalesApi } from '../services/stats-wales-api';
...
-                    tag: {
-                        name: 'errors.upload.no-csv-data',
-                        params: {}
-                    }
-                }
-            ]
-        };
-        res.status(400);
-        res.render('publish/upload', err);
+        generateFileError(req, res);
```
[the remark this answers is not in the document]
j-maynard: Because it creates the error and sets the return status and the render.  So if you do this you just need to return... That could be clearer in the function name.  So no return object... just create an error and pass it back to the view.

src/routes/publish.ts
```diff
@@ -8,196 +8,602 @@ import { StatsWalesApi } from '../services/stats-wales-api';
...
+        logger.debug('Current title was missing from the session.  Something might have gone wrong');
+        req.session.errors = generateViewErrors(undefined, 400, [generateError('title', 'errors.title_missing', {})]);
+        res.redirect(
+            `/${lang}/${req.i18n.t('routes.publish.start', { lng: lang })}/${t('routes.publish.title', { lng: lang })}`
+        );
+        return;
+    }
+    const file = req.file;
+    if (!file) {
+        generateFileError(req, res);
```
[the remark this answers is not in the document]
j-maynard: Same as above... I think this came about as an attempt to reduce code duplication.

src/routes/publish.ts
```diff
@@ -8,196 +8,602 @@ import { StatsWalesApi } from '../services/stats-wales-api';
...
+}
 
-    const processedCSV = await statsWalesApi.uploadCSV(fileData, fileName, title);
+function cleanupSession(req: AuthedRequest) {
+    req.session.currentDataset = undefined;
+    req.session.currentRevision = undefined;
+    req.session.currentImport = undefined;
+    req.session.dimensionCreationRequest = undefined;
+    req.session.errors = undefined;
+    req.session.save();
```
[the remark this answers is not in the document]
j-maynard: I considered that but if we move the users JWT to the session as well for example destroying it might not be a good idea.  Might be worth changing the name as so far this relates specifically to the create journey.  If things go wrong we return to the start, give the errors to the view and then empty the things we need to empty when the user starts a new journey.

src/routes/publish.ts
```diff
@@ -8,196 +8,602 @@ import { StatsWalesApi } from '../services/stats-wales-api';
...
-        logger.error('No dataset in session');
-        res.redirect(`/${req.i18n.language}/publish/title`);
+    const currentDataset = checkCurrentDataset(req, res);
+    if (!currentDataset) {
+        return;
+    }
+
+    const currentRevision = checkCurrentRevision(req, res);
+    if (!currentRevision) {
+        return;
```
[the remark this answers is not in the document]
j-maynard: It typically returns a 500 error and redirects the user back to the start of the publish journey.  This is because something has gone wrong with their session.  Also again typescript so you need to make sure the thing exists before you can use it.

src/routes/publish.ts
```diff
@@ -8,196 +8,602 @@ import { StatsWalesApi } from '../services/stats-wales-api';
...
+            generateError('confirm', 'errors.preview.remove_error', {})
+        ]);
+        req.session.save();
+        res.redirect(
+            `/${lang}/${req.i18n.t('routes.publish.start', { lng: lang })}/${req.i18n.t('routes.publish.preview', { lng: lang })}`
+        );
+    }
+}
+
+publish.post('/preview', upload.none(), async (req: AuthedRequest, res: Response) => {
```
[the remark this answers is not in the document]
j-maynard: It means that no file is expect as part of the form data... Its part of multer allowing us to read for data from the frontend as well as allowing us to receive files.

src/routes/publish.ts
```diff
@@ -8,196 +8,602 @@ import { StatsWalesApi } from '../services/stats-wales-api';
...
+
+    const currentRevision = checkCurrentRevision(req, res);
+    if (!currentRevision) {
+        return;
+    }
+
+    const currentFileImport = checkCurrentFileImport(req, res);
+    if (!currentFileImport) {
+        return;
+    }
```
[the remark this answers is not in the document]
j-maynard: So I think you have to have the checks because its TypeScript... If the object could be undefined (which it can be when getting it from the session) you have to confirm the object you want to use actually exists otherwise the typescript compiler will throw errors at you.

src/services/stats-wales-api.ts
```diff
@@ -179,14 +174,165 @@ export class StatsWalesApi {
...
+                        }
+                    ],
+                    dataset: undefined
+                } as UploadErrDTO;
+            });
+        return processedCSV;
+    }
+
+    public async removeFileImport(datasetId: string, revisionId: string, importId: string) {
+        const revisiedDataset = await fetch(
```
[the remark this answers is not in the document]
j-maynard: Yup... Changed to `revisedDataset`

src/views/publish/sources.ejs
j-maynard: Because we haven't done the bigger bit of work of ripping the GEL out of the prototype.  So it was just to make it look like the screenshot.  When we bring the GEL over we'll either need to include this in the CSS or hopefully it'll be covered by what we have already.

src/views/publish/start.ejs
j-maynard: Added the new line... No typically I don't have the add a newline option turned on as I've found there are more times than not that I don't want/need a new line at the end of a file.

test/publish.test.ts
j-maynard: Fixed

tools/copyAssets.ts
j-maynard: Removed the copy to root... That was a mistake when I was debugging an issue with translations which was affecting a few things.

src/views/partials/header.ejs
wheelsandcogs: Sorry I was getting it confused with the existing gov phase banner for the service which looks like it's already there.

In that case it feels a bit excessive to call certain pages when literally the entire service is still in active development and subject to change, but I'm not willing to die on that hill if people want it 😄

src/routes/publish.ts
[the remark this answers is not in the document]
wheelsandcogs: Ah, was getting confused with `generateError`

src/routes/publish.ts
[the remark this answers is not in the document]
wheelsandcogs: Is this required on any post route that does not receive files?

src/routes/publish.ts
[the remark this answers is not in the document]
wheelsandcogs: That can be resolved with typing - before the middleware is run your session type props are optional. Then the middleware throw exceptions if they're missing. Once the middleware has run, you use a type that removes optionality from those props.

wheelsandcogs (approved):
🚢

PR #13, opened 2024-09-19
Implement Redis Session Storage

j-maynard:
Implements redis for session storage.  Pipeline needs to be updated with a new `REDIS_URL` and `REDIS_ACCESS_KEY` variables to run.

### DIFF
--- a/test/app.test.ts
+++ b/test/app.test.ts
@@ -4,8 +4,6 @@ import request from 'supertest';
 import { Request, Response, NextFunction } from 'express';
 import { http, HttpResponse } from 'msw';
 import { setupServer } from 'msw/node';
-import { CookieAccessInfo } from 'cookiejar';
-import supertest from 'supertest';
 
 import { ENGLISH, WELSH, i18next } from '../src/middleware/translation';
 import app from '../src/app';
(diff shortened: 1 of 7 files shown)


test/app.test.ts
```diff
@@ -411,4 +439,35 @@ describe('Test app.ts', () => {
...
+        expect(res.text).toContain(`<td class="govuk-table__cell">test1</td>`);
+        expect(res.text).toContain(`<td class="govuk-table__cell">3423196</td>`);
+        expect(res.text).toContain(`<td class="govuk-table__cell">2001-09-20</td>`);
+        // Last Row
+        expect(res.text).toContain(`<td class="govuk-table__cell">2</td>`);
+        expect(res.text).toContain(`<td class="govuk-table__cell">AcHVoWJblA</td>`);
+        expect(res.text).toContain(`<td class="govuk-table__cell">4470652</td>`);
+        expect(res.text).toContain(`<td class="govuk-table__cell">2002-03-18</td>`);
+        expect(res.text).toContain('Showing rows 1 - 2 of 2');
+    });
```
wheelsandcogs, line 472: I'd probably avoid testing the exact html that comes back and just make sure the data is present - otherwise if we change the classes etc it will break the test.

wheelsandcogs (approved):
I've got some tweaks I'd like to make around the session config but I can make those changes in my config PR once I've rebased.

test/app.test.ts
wheelsandcogs: We can resolve this in another pass though. :ship:

PR #14, opened 2024-09-24
Minor broken link in the view and an issue when there are no datasets

j-maynard:
Fixes the broken link on the home page.  Fixes a small bug where you might not have a file list array before you look at its length.  Fixes the error this creates.

### DIFF
--- a/src/views/view/list.ejs
+++ b/src/views/view/list.ejs
@@ -4,7 +4,7 @@
         <!-- <a href="#" class="govuk-back-link">Back</a> -->
         <main class="govuk-main-wrapper" id="main-content" role="main">
             <h1 class="govuk-heading-xl"><%= t('view.list.heading') %></h1>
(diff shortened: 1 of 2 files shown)


src/views/view/list.ejs
```diff
@@ -4,7 +4,7 @@
         <!-- <a href="#" class="govuk-back-link">Back</a> -->
         <main class="govuk-main-wrapper" id="main-content" role="main">
             <h1 class="govuk-heading-xl"><%= t('view.list.heading') %></h1>
-            <% if (locals?.filelist.length > 0) {%>
+            <% if (locals?.filelist && locals?.filelist.length > 0) {%>
```
wheelsandcogs, line 7:
```suggestion
            <% if (locals?.filelist?.length > 0) {%>
```

wheelsandcogs (approved):
🚢

PR #17, opened 2024-09-27
Update the frontend to match the backend

wheelsandcogs:
Various updates to sync the frontend with the backend changes in https://github.com/Marvell-Consulting/statswales-backend/pull/18
 * use a similar config system to the backend
 * update eslint and use the same rules
 * update DTOs to match backend
 * use consistent filenaming
 * bump dependencies where possible

NOTE: `msw` has been locked at `2.4.3` for the time being as `2.4.4` onwards causes issues with `supertest.attach()`.


j-maynard (approved):
LGTM 🚀

PR #18, opened 2024-10-01
Fixes missing strings, the odd bug and tasklist

j-maynard:
Following feedback from Nicole I've updated the task list with status information and left a comment on the function to generate the tasklist to generate this on the backend.

I've checked for missing error strings and added these in to the english translations file.  The view dataset page has been tweaked so we can download the CSV files from the imports.  This is useful for creating a better testing environment and will proves out that we can easily download CSVs from the Datalake via the backend to the frontend.


src/dtos2/single-language/dataset.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export interface SingleLanguageDataset {
+    id: string;
+    creation_date: string;
+    created_by: string;
+    live?: string;
+    archive?: string;
+    dimensions?: SingleLanguageDimension[];
+    revisions: RevisionDTO[];
+    datasetInfo?: DatasetInfoDTO;
```
wheelsandcogs, line 13: This is fine for now, but I think the "standard" use case is probably always single language, and the "special" use case would be to return multiple languages?

e.g. I'm nearly always in either en or cy, the only time I might want both at the same time is viewing or exporting the entire nested object?

wheelsandcogs:
Closing for weirdness

PR #20, opened 2024-10-02
Fix bugs

wheelsandcogs:
Following feedback from Nicole I've updated the task list with status information and left a comment on the function to generate the tasklist to generate this on the backend.

I've checked for missing error strings and added these in to the english translations file. The view dataset page has been tweaked so we can download the CSV files from the imports. This is useful for creating a better testing environment and will proves out that we can easily download CSVs from the Datalake via the backend to the frontend.


j-maynard (approved):
LGTM 🚀

PR #22, opened 2024-10-03
SW-209: move language to header for api requests

wheelsandcogs:
This is a companion PR to https://github.com/Marvell-Consulting/statswales-backend/pull/29

This updates the `StatsWalesApi` service to
 * send language as an `Accept-Language` header instead of the path
 * throw exceptions on error instead of returning wrapped objects


src/config/app-config.interface.ts
```diff
@@ -31,6 +31,7 @@ export interface AppConfig {
     auth: {
         jwt: {
             secret: string;
+            cookieDomain: string;
```
wheelsandcogs, line 34: This was required to get the jwt cookie to delete properly.

src/dtos/confirmed-import-dto.ts
```diff
@@ -1,13 +0,0 @@
...
-export interface ConfirmedImportDTO {
-    success: boolean;
-    fileImport: FileImportDTO;
-}
-
-export interface UploadErrDTO {
-    success: boolean;
-    fileImport: FileImportDTO | undefined;
-    errors: ViewError[];
-}
```
wheelsandcogs, line 13: These should no-longer be required as the api service will throw exceptions instead.

src/middleware/ensure-authenticated.ts
```diff
@@ -38,7 +38,7 @@ export const ensureAuthenticated: RequestHandler = (req: AuthedRequest, res, nex
     } catch (err) {
         logger.error(`authentication failed: ${err}`);
         res.status(401);
-        return res.redirect('/auth/login');
+        return res.redirect(`/${req.language}/auth/login`);
```
wheelsandcogs, line 41: auth routes now have language in the path.

src/routes/auth.ts
```diff
@@ -21,12 +24,12 @@ auth.get('/login', (req: Request, res: Response) => {
...
 auth.get('/google', (req: Request, res: Response) => {
     logger.debug('Sending user to backend for google authentication');
-    res.redirect(`${config.backend.url}/auth/google`);
+    res.redirect(`${config.backend.url}/auth/google?lang=${req.language}`);
 });
 
 auth.get('/onelogin', (req: Request, res: Response) => {
     logger.debug('Sending user to backend for onelogin authentication');
-    res.redirect(`${config.backend.url}/auth/onelogin`);
+    res.redirect(`${config.backend.url}/auth/onelogin?lang=${req.language}`);
```
wheelsandcogs, line 32: the lang query param doesn't do anything yet, but we'll need to pass the language to the backend when triggering the auth flow so that we can come back in the same language.

Currently you will be returned to the frontend in whatever language your browser is set to.

src/routes/auth.ts
```diff
@@ -53,10 +56,11 @@ auth.get('/callback', (req: Request, res: Response) => {
...
-    res.redirect('/');
+    res.redirect(`/${req.language}`);
 });
 
-auth.get('/logout', (req: Request, res: Response) => {
-    res.clearCookie('jwt');
-    res.redirect('/auth/login');
+auth.get('/logout', (req: AuthedRequest, res: Response) => {
+    logger.debug('logging out user');
+    res.clearCookie('jwt', { domain: cookieDomain });
```
wheelsandcogs, line 64: the auth cookie previously wasn't clearing properly on logout - setting the domain seems to fix it.

src/routes/error-handler.ts
```diff
@@ -0,0 +1,17 @@
...
+            res.status(404);
+            res.render('errors/not-found');
+            break;
+        case 500:
+        default:
+            res.status(500);
+            res.render('errors/server-error');
+            break;
+    }
+};
```
wheelsandcogs, line 17: Fall-through error handler for anything that's not handled by the route.

src/routes/publish.ts
```diff
@@ -120,62 +118,51 @@ function checkCurrentFileImport(req: AuthedRequest, res: Response): FileImportDT
...
 }
 
-function handleProcessedCSV(processedCSV: UploadDTO | UploadErrDTO, req: AuthedRequest, res: Response) {
-    const lang = req.i18n.language;
-    if (!processedCSV.success) {
-        const err = processedCSV as UploadErrDTO;
-        req.session.errors = generateViewErrors(err.dataset?.id, 500, err.errors);
-        res.redirect(`/${lang}/${req.i18n.t('routes.publish.start', { lng: lang })}/`);
-        return;
-    }
```
wheelsandcogs, line 136: Rather than wrapping the API return in an object with a success/fail flag, it will now always return the actual DTO we're interested in on success, and throw an exception on fail.

src/routes/publish.ts
```diff
@@ -120,62 +118,51 @@ function checkCurrentFileImport(req: AuthedRequest, res: Response): FileImportDT
         req.session.errors = generateViewErrors(undefined, 500, [
             generateError('session', 'errors.session.current_import_missing', {})
         ]);
-        res.redirect(`/${lang}/${req.i18n.t('routes.publish.start', { lng: lang })}/`);
+        res.redirect(`/${lang}/${req.i18n.t('routes.publish.start', { lng: lang })}`);
```
wheelsandcogs, line 121: They'll be a URL builder coming in a subsequent PR that will handle this a bit nicer. Something like:

```
res.redirect(getUrl('/path/to/page', { lang }));
```

j-maynard (approved):
Single comment... Happy to leave it for follow up unless you want to fix it now and then I can reapprove.

src/views/errors/server-error.ejs
```diff
@@ -0,0 +1,10 @@
+<%- include("../partials/header"); %>
+
+<div class="govuk-width-container app-width-container">
+  <!-- <a href="#" class="govuk-back-link">Back</a> -->
+  <main class="govuk-main-wrapper" id="main-content" role="main">
+      <h1 class="govuk-heading-xl"><%- t('errors.server_error') %></h1>
+  </main>
```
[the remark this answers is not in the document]
wheelsandcogs: 100% - we can add whatever we like here, will fill it out a bit more in subsequent PRs.

PR #23, opened 2024-10-07
Various language bug fixes

wheelsandcogs:
This fixes the following issues around language switching:

### BUG - SW-199: Website goes back to English when Welsh has been selected
FIX - We now cache the selected language in a "lang" cookie so if a URL is hit without a locale in the path, we use the cookie lang if available before defaulting to browser lang.

### BUG - SW-200: Site goes back to English when homepage is selected
FIX - The language switching now uses a query param and will work on any page.

### BUG - SW-195: Language button doen't work and then inverts itself
FIX - The language switching link is always in the alternate language


easternbloc (approved):
LGTM!

PR #24, opened 2024-10-10
SW-241: Log the user out if the frontend receives a 401 response from the API

wheelsandcogs:
The user's JWT token will now be deleted if the backend sends a 401 response.

Previously they would still appear logged in but every request for a secured endpoint would fail.


j-maynard (approved):
LGTM 🚀

PR #25, opened 2024-10-11
Pass 401 exceptions through to the default error handler

wheelsandcogs:
There were a number of routes that do their own error handling so we need to check for 401 and then hand it off to the default error handler for the logout.


j-maynard (approved):
Approved but we'll need to do some work on the use of those `any` errors. 🙂

PR #29, opened 2024-10-16
Add services to request via middleware

wheelsandcogs:
Currently if we want to use the StatsWalesApi anywhere we have to create a new instance of it in each router, as it needs the request object for the current language and the jwt token.

There's a pattern in Express where services (e.g. i18next, session, etc) get added to the request object so that they're available for the lifetime of the request.

I've basically done that with the StatsWalesApi which is now available via `req.swapi` and a locale based URL builder `req.buildUrl`.

I've also extended the standard request object with `req.user` and `req.jwt` so that we don't have to type the request object as `AuthedRequest` any more.


j-maynard (approved):
LGTM 🚀

PR #30, opened 2024-10-16
Use url builder to generate all URLs

wheelsandcogs:
This simplifies creating URLs by not having to worry about translation - just pass the builder the actual route and the expected locale, and it will return a URL string with the locale and a translated path.

If the current locale is English, it will just skip the path translation completely to save on the translation lookups.

### DIFF
--- a/src/middleware/translations/en.json
+++ b/src/middleware/translations/en.json
@@ -214,18 +214,15 @@
     "routes": {
         "healthcheck": "healthcheck",
         "feedback": "feedback",
-        "view": {
-            "start": "dataset"
-        },
-        "publish": {
-            "start": "publish",
-            "title": "title",
-            "upload": "upload",
-            "preview": "preview",
-            "confirm": "confirm",
-            "sources": "sources",
-            "source_confirmation": "source-confirmation",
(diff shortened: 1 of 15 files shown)


src/middleware/translations/en.json
```diff
@@ -211,18 +211,15 @@
...
+        "dataset": "dataset",
+        "publish": "publish",
+        "start": "start",
+        "title": "title",
+        "preview": "preview",
+        "upload": "upload",
+        "confirm": "confirm",
+        "sources": "sources",
+        "source_confirmation": "source-confirmation",
+        "tasklist": "tasklist"
```
wheelsandcogs, line 223: Technically these should never get used as the url builder skips attempting to translate URLs when the locale is English.

However, we'll need them as fallback if the Welsh is missing, and the Welsh language translations need the right keys in order to work, so it makes sense for the English translation file to have them as a source.

wheelsandcogs: PS - these must be flat - the builder translates each bit of the path individually not as a whole.

j-maynard (approved):

PR #32, opened 2024-10-17
Use unnamed col fallback on sources as well

wheelsandcogs:
This adds the same placeholder text for missing column headers on the sources page (previously only the preview was showing it).

Screenshot 2024-10-17 at 09 52 16


j-maynard (approved):
LGTM 🚀

PR #46, opened 2024-11-26
A couple more fixes for data-providers

wheelsandcogs:
Fixes:
 * should be a line break gap between the instruction text and the hint text 
 * Radio and continue button are out of alignment with the parts above
 * when you select 'No specific source from data provider' it should populate the summary table with "No specific source from data provider"


j-maynard (approved):
LGTM 🚀

PR #48, opened 2024-11-26
Refactor the frontend to support the changes to the backend

j-maynard:
The backend refactor resulted in a number of changes to the frontend mostly around changing `imports` to `fact-tables`.  The update adds support for line numbers in csv table previews.  It also adds support for proper GDS style pagination which resulted in the preview and view pages needing to have their layouts adjusted.  This will need to be cleared by UX and content.  It does however bring these pages much more in line with the GDS Design System and service standard.


src/controllers/publish.ts
```diff
@@ -104,27 +134,67 @@ export const uploadFile = async (req: Request, res: Response, next: NextFunction
...
+    if (highlight + 1 < end) {
+        sequence.push(highlight + 1);
+        sequence.push('...', end);
+    } else {
+        // eslint-disable-next-line @typescript-eslint/naming-convention
+        sequence.push(...Array.from({ length: end - highlight }, (_, index) => highlight + 1 + index));
+    }
+
+    return sequence;
+}
```
wheelsandcogs, line 168: Can we stick this in `/src/utils/pagination` please? Just thinking it might be useful for routes outside the publish journey.

src/dtos/dimension.ts
```diff
@@ -1,13 +1,14 @@
...
     type: string;
-    start_revision_id: string;
-    finish_revision_id?: string;
-    validator?: string;
-    sources?: SourceDTO[];
+    extractor?: object;
+    joinColumn?: string; // <-- Tells you have to join the dimension to the fact_table
+    factTableColumn: string; // <-- Tells you which column in the fact table you're joining to
+    isSliceDimension: boolean;
+    lookupTable?: LookupTableDTO;
```
wheelsandcogs, line 12: If we're sticking to the convention of DTO props always being snake_case then we'll want to tidy this up at some point but not a blocker here.

src/routes/dataset.ts
```diff
@@ -7,12 +7,45 @@ import { fetchDataset } from '../middleware/fetch-dataset';
...
+    if (highlight + 1 < end) {
+        sequence.push(highlight + 1);
+        sequence.push('...', end);
+    } else {
+        // eslint-disable-next-line @typescript-eslint/naming-convention
+        sequence.push(...Array.from({ length: end - highlight }, (_, index) => highlight + 1 + index));
+    }
+
+    return sequence;
+}
```
wheelsandcogs, line 47: Ah - so this is actually reused, would be good to extract this to utils

src/services/stats-wales-api.ts
```diff
@@ -114,70 +114,74 @@ export class StatsWalesApi {
...
         pageSize: number
     ): Promise<ViewDTO> {
         logger.debug(
-            `Fetching preview for dataset: ${datasetId}, revision: ${revisionId}, import: ${importId}, page: ${pageNumber}, pageSize: ${pageSize}`
+            `Fetching preview for dataset: ${datasetId}, revision: ${revisionId}, import: ${factTableId}, page: ${pageNumber}, pageSize: ${pageSize}`
         );
 
         return this.fetch({
-            url: `dataset/${datasetId}/revision/by-id/${revisionId}/import/by-id/${importId}/preview?page_number=${pageNumber}&page_size=${pageSize}`
+            url: `dataset/${datasetId}/revision/by-id/${revisionId}/fact-table/by-id/${factTableId}/preview?page_number=${pageNumber}&page_size=${pageSize}`
```
wheelsandcogs, line 171: Do we think this `by-id` stuff is still valuable since the schema change? If so, fine, otherwise let's strip it out in a future refactor.

src/controllers/publish.ts
```diff
@@ -85,11 +87,39 @@ export const uploadFile = async (req: Request, res: Response, next: NextFunction
...
+            } else if (req.file.mimetype === 'application/x-gzip') {
+                const ext = req.file.originalname.split('.').reverse()[1];
+                switch (ext) {
+                    case 'json':
+                    case 'csv':
+                        break;
+                    default:
+                        throw new Error(`unsupported format ${ext}`);
+                }
+            }
```
wheelsandcogs, line 122: I'd probably extract this to a util function but not a blocker.

wheelsandcogs (approved):
There's a couple of functions I'd stick in their own files but otherwise looks fine.

src/services/stats-wales-api.ts
j-maynard: I started sketching out the API on the my miro board the other day... and came up with this.  For revisions you'll want by-index/version and ID.  Dataset you'll probably want to get by URLized title or code, we've not talked about code recently...  but its the codes like `tran0303`... That'll play into "Don't break the internet"... I believe its probably still a thing, I suspect the sticking point will be how those codes are generated...  While I'd like a simpler route at least `by-id` explains what you're doing.

If you want to suggest an alternative route layout I'm all for it.  I have other ideas but until MVP lands I think its something we should stick in the tech-debt pile to talk about and agree later...  I'd even put a date on it of the 1st of February to talk about it and agree which way we want to go.

src/controllers/publish.ts
j-maynard: I'll land this PR and raise a small one to extract and reuse.

src/routes/dataset.ts
j-maynard: Will extract and reuse in follow up PR

src/controllers/publish.ts
j-maynard: I'll do it with pagination.

PR #50, opened 2024-11-28
Fix paging bug

j-maynard:
Fixes a few bugs in the pagination code.  Moves reusable code out to
utilities as per agreement in previous PR.  Fixes some minor issues with
preview and data views.

The developer view now introduces a possible table component allowing
for overflow on x and y for big tables and introduces sticky table
headers to make viewing the tables easier.  Should be tested with users
and user research later.


wheelsandcogs (approved):
:shipit:

PR #59, opened 2024-12-16
Fixes issues where the wrong values are encoded in frontend

j-maynard:
Fixes an issue where the wrong values are encoded in the frontend.  Its trying to set specific point in time formats as all caps when they shouldn't be.

### DIFF
--- a/src/views/publish/point-in-time-chooser.ejs
+++ b/src/views/publish/point-in-time-chooser.ejs
@@ -21,7 +21,7 @@
                         <fieldset class="govuk-fieldset">
                             <div class="govuk-radios" data-module="govuk-radios">
                                 <div class="govuk-radios__item">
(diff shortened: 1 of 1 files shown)


src/views/publish/point-in-time-chooser.ejs
```diff
@@ -48,7 +48,7 @@
                                     </div>
                                 </div>
                                 <div class="govuk-radios__item">
-                                    <input class="govuk-radios__input" id="quarter-format-5" name="dateFormat" type="radio" value="YYYYMMDD">
+                                    <input class="govuk-radios__input" id="quarter-format-5" name="dateFormat" type="radio" value="yyyyMMd d">
```
wheelsandcogs, line 51: is the space meant to be there?

j-maynard: Good catch... no its not

j-maynard: Fixed!

PR #66, opened 2025-01-03
Implement measures and preview

j-maynard:
This PR implements measures and data cube previews in the publisher journey.  It adds the ability to download the cube in multiple formats including CSV, Excel and Parquet as well as downloading the raw cube in duckdb format.

Measures are added as a look up table upload in the measures lookup table format.

The preview page currently lacks the topics associated with the cube due to a bug on the backend which doesn't currently pass the topics through in a user readable format.

### DIFF
--- a/src/controllers/publish.ts
+++ b/src/controllers/publish.ts
@@ -395,9 +395,9 @@ export const measurePreview = async (req: Request, res: Response, next: NextFunc
                 const error = err as ApiException;
                 logger.debug(`Error is: ${JSON.stringify(error, null, 2)}`);
                 if (error.status === 400) {
-                    res.status(400);
                     logger.error('Measure lookup table did not match data in the fact table.', err);
                     const failurePreview = JSON.parse(error.body as string) as ViewErrDTO;
+                    res.status(400);
                     res.render('publish/measure-match-failure', {
                         ...failurePreview,
                         measure
--- a/src/controllers/publish.ts
+++ b/src/controllers/publish.ts
@@ -37,7 +37,7 @@ import { ViewError } from '../dtos/view-error';
 import { logger } from '../utils/logger';
 import { ViewDTO, ViewErrDTO } from '../dtos/view-dto';
 import { SourceType } from '../enums/source-type';
-import { FactTableInfoDto } from '../dtos/fact-table-info';
+import { FactTableInfoDTO } from '../dtos/fact-table-info';
 import { SourceAssignmentDTO } from '../dtos/source-assignment-dto';
 import { UnknownException } from '../exceptions/unknown.exception';
 import { TaskListState } from '../dtos/task-list-state';
@@ -59,7 +59,7 @@ import { nestTopics } from '../utils/nested-topics';
 import { OrganisationDTO } from '../dtos/organisation';
 import { TeamDTO } from '../dtos/team';
 import { DimensionType } from '../enums/dimension-type';
-import { DimensionPatchDto } from '../dtos/dimension-patch-dto';
+import { DimensionPatchDTO } from '../dtos/dimension-patch-dto';
 import { ApiException } from '../exceptions/api.exception';
 import { DimensionInfoDTO } from '../dtos/dimension-info';
 import { YearType } from '../enums/year-type';
@@ -152,7 +152,7 @@ export const factTablePreview = async (req: Request, res: Response, next: NextFu
 
     // if sources have previously been assigned a type, this is a revisit
     const revisit =
-        factTable.fact_table_info?.filter((factTableInfo: FactTableInfoDto) => Boolean(factTableInfo.column_type))
+        factTable.fact_table_info?.filter((factTableInfo: FactTableInfoDTO) => Boolean(factTableInfo.column_type))
             .length > 0;
 
     if (req.method === 'POST') {
@@ -193,7 +193,7 @@ export const factTablePreview = async (req: Request, res: Response, next: NextFu
 export const sources = async (req: Request, res: Response, next: NextFunction) => {
     const { dataset, revision, factTable } = res.locals;
     const revisit =
-        factTable.fact_table_info?.filter((factTableInfo: FactTableInfoDto) => Boolean(factTableInfo.column_type))
+        factTable.fact_table_info?.filter((factTableInfo: FactTableInfoDTO) => Boolean(factTableInfo.column_type))
             .length > 0;
     let error: ViewError | undefined;
     let errors: ViewError[] | undefined;
@@ -208,7 +208,7 @@ export const sources = async (req: Request, res: Response, next: NextFunction) =
         if (req.method === 'POST') {
             const counts = { unknown: 0, dataValues: 0, footnotes: 0, measure: 0 };
             const sourceAssignment: SourceAssignmentDTO[] = factTable.fact_table_info.map(
-                (factTableInfo: FactTableInfoDto) => {
+                (factTableInfo: FactTableInfoDTO) => {
                     const sourceType = req.body[`column-${factTableInfo.column_index}`];
                     if (sourceType === SourceType.Unknown) counts.unknown++;
                     if (sourceType === SourceType.DataValues) counts.dataValues++;
@@ -453,10 +453,10 @@ export const measureReview = async (req: Request, res: Response, next: NextFunct
         if (req.method === 'POST') {
             logger.debug(`User has reviewed measure lookup table.`);
             switch (req.body.confirm) {
-                case 'true':
+                case 'continue':
                     res.redirect(req.buildUrl(`/publish/${dataset.id}/tasklist`, req.language));
                     break;
-                case 'goback':
+                case 'cancel':
                     try {
                         await req.swapi.resetMeasure(dataset.id);
                         res.redirect(req.buildUrl(`/publish/${dataset.id}/measure`, req.language));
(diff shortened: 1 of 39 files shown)


src/controllers/publish.ts
```diff
@@ -295,10 +279,223 @@ export const taskList = async (req: Request, res: Response, next: NextFunction)
...
+    const fileStream = await req.swapi.getRevisionCube(dataset.id, revision.id);
+    res.writeHead(200, {
+        // eslint-disable-next-line @typescript-eslint/naming-convention
+        'Content-Type': 'application/octet-stream',
+        // eslint-disable-next-line @typescript-eslint/naming-convention
+        'Content-disposition': `attachment;filename=${revision.id}.duckdb`
+    });
+    const readable: Readable = Readable.from(fileStream);
+    readable.pipe(res);
+};
```
wheelsandcogs, line 364: As with the backend it feels like we could slim this down to one request handler and one api method e.g. something like

```ts
export const download = async (req: Request, res: Response, next: NextFunction) => {
    const dataset = singleLangDataset(res.locals.dataset, req.language);
    const revision = res.locals.revision;
    const format = req.query.format;

    const fileStream = await req.swapi.getRevisionCube(dataset.id, revision.id, format);
    res.writeHead(200, getContentHeaders(format));
    const readable: Readable = Readable.from(fileStream);
    readable.pipe(res);
};
```

src/controllers/publish.ts
```diff
@@ -295,10 +279,223 @@ export const taskList = async (req: Request, res: Response, next: NextFunction)
...
+
 export const redirectToTasklist = (req: Request, res: Response) => {
     res.redirect(req.buildUrl(`/publish/${req.params.datasetId}/tasklist`, req.language));
 };
 
+export const measurePreview = async (req: Request, res: Response, next: NextFunction) => {
+    try {
+        const dataset = singleLangDataset(res.locals.dataset, req.language);
+        const measure = singleLangDataset(res.locals.dataset, req.language).measure;
```
wheelsandcogs, line 373:
```suggestion
        const measure = dataset.measure;
```

src/controllers/publish.ts
```diff
@@ -295,10 +279,223 @@ export const taskList = async (req: Request, res: Response, next: NextFunction)
...
+        }
+        let errors: ViewErrDTO | undefined;
+
+        if (req.method === 'POST') {
+            logger.debug(`User has reviewed measure lookup table.`);
+            switch (req.body.confirm) {
+                case 'true':
+                    res.redirect(req.buildUrl(`/publish/${dataset.id}/tasklist`, req.language));
+                    break;
+                case 'goback':
```
wheelsandcogs, line 459: Tiny thing but just for consistency's sake I'd suggest either `true` & `false` or maybe `continue` & `cancel`.

src/controllers/publish.ts
```diff
@@ -295,10 +279,223 @@ export const taskList = async (req: Request, res: Response, next: NextFunction)
...
+            return;
+        }
+
+        const dataPreview = await req.swapi.getMeasurePreview(res.locals.dataset.id);
+        if (errors) {
+            res.status(errors.status || 500);
+            res.render('publish/measure-review', { ...dataPreview, measure, review: true, errors });
+        } else {
+            res.render('publish/measure-review', { ...dataPreview, measure, review: true });
+        }
```
wheelsandcogs, line 492: The pages (should hopefully) only render errors when set, so I'd just always pass the `errors` prop and save a few lines...

```
if (errors) {
  res.status(errors.status || 500);
}
res.render('publish/measure-review', { ...dataPreview, measure, review: true, errors });
```

src/dtos/lookup-table.ts
```diff
@@ -1,4 +1,4 @@
-export class LookupTableDTO {
+export class LookupTable {
```
wheelsandcogs, line 1: Any reason for breaking with the convention here by removing the DTO suffix? I'm not precious about keeping it but we should either always have it or never have it in that case :)

src/dtos/measure.ts
```diff
@@ -0,0 +1,11 @@
+import { LookupTable } from './lookup-table';
+import { MeasureInfo } from './measure-info';
+
+export class Measure {
+    id: string;
+    dataset_id: string;
+    fact_table_column: string;
+    join_column: string | null;
+    lookup_table?: LookupTable;
+    measure_info: MeasureInfo[] | undefined;
```
wheelsandcogs, line 10:
```suggestion
    measure_info?: MeasureInfo[];
```

src/middleware/services.ts
```diff
@@ -18,6 +18,8 @@ export const initServices = (req: Request, res: Response, next: NextFunction): v
         res.locals.referrer = req.get('Referrer');
         res.locals.parseISO = parseISO;
         res.locals.dateFormat = format;
+        res.locals.dateAdd = add;
+        res.locals.languague = req.language;
```
wheelsandcogs, line 22: typo - but do we actually need this? If it's for the templates then it already exists on `i18n.language` in the views.

src/routes/publish.ts
```diff
@@ -64,6 +71,19 @@ publish.post('/:datasetId/sources', fetchDataset, upload.none(), sources);
 /* Tasklist */
 publish.get('/:datasetId/tasklist', fetchDataset, taskList);
 
+/* Cube Preview */
+publish.get('/:datasetId/cube-preview', fetchDataset, cubePreview);
+publish.get('/:datasetId/cube-preview/csv', fetchDataset, downloadAsCSV);
+publish.get('/:datasetId/cube-preview/parquet', fetchDataset, downloadAsParquet);
+publish.get('/:datasetId/cube-preview/excel', fetchDataset, downloadAsExcel);
+publish.get('/:datasetId/cube-preview/cube', fetchDataset, downloadAsDuckDb);
```
wheelsandcogs, line 79: As mentioned in the controllers, this could probably just be 

```
publish.get('/:datasetId/cube-preview', fetchDataset, cubePreview);
```

with URLs like `/:datasetId/cube-preview?format=csv` etc

src/services/stats-wales-api.ts
```diff
@@ -149,6 +169,38 @@ export class StatsWalesApi {
...
+        }).then((response) => response.body as ReadableStream);
+    }
+
+    public async getRevisionCube(datasetId: string, revisionId: string): Promise<ReadableStream> {
+        logger.debug(`Fetching CSV for revision: ${revisionId}...`);
+
+        return this.fetch({
+            url: `dataset/${datasetId}/revision/by-id/${revisionId}/cube`
+        }).then((response) => response.body as ReadableStream);
+    }
```
wheelsandcogs, line 202: As above, this could probably be 

```ts
public async getRevisionCube(datasetId: string, revisionId: string, format: FileFormat): Promise<ReadableStream> {
    logger.debug(`Fetching CSV for revision: ${revisionId}...`);

    return this.fetch({
        url: `dataset/${datasetId}/revision/by-id/${revisionId}/cube?format=${format}`
    }).then((response) => response.body as ReadableStream);
}
```

src/views/publish/cube-preview.ejs
```diff
@@ -0,0 +1,291 @@
...
+                <%= t('publish.cube_preview.panel') %>
+            </p>
+        </div>
+
+        <h2 class="govuk-heading-l">
+            <%= t('dataset_view.key_information') %>
+        </h2>
+
+        <%
+            const publishDate = new Date(Math.max.apply(null, dataset.revisions.map((rev) => new Date(Date.parse(rev.publish_at? rev.publish_at : rev.created_at)))));
```
wheelsandcogs, line 19: Would be better to extract the code inside `new Date(...)` to a function and name it descriptively so it's self-documenting, and also just pass a publish date through to the template rather than doing the work here.

src/views/publish/cube-preview.ejs
```diff
@@ -0,0 +1,291 @@
...
+            if (dataset.datasetInfo.update_frequency) {
+                if (dataset.datasetInfo.update_frequency.frequency_unit === 'year')
+                    updateObj = { years: dataset.datasetInfo.update_frequency.frequency_value };
+                if (dataset.datasetInfo.update_frequency.frequency_unit === 'month')
+                    updateObj = { months: dataset.datasetInfo.update_frequency.frequency_value };
+                if (dataset.datasetInfo.update_frequency.frequency_unit === 'week')
+                    updateObj = { weeks: dataset.datasetInfo.update_frequency.frequency_value };
+                if (dataset.datasetInfo.update_frequency.frequency_unit === 'day')
+                    updateObj = { days: dataset.datasetInfo.update_frequency.frequency_value };
+            }
```
wheelsandcogs, line 30: Same here - don't do calculations in the template, just pass `nextUpdate` through from the route handler.

It can probably just be 

```ts
const update = dataset.datasetInfo.update_frequency;
const nextUpdate = update ? dateAdd(publishDate, { [update.frequency_unit]: update.frequency_value }) : undefined;
```

wheelsandcogs (approved):
Few bits and bobs that might be worth sorting now, nothing major

src/middleware/services.ts
j-maynard: Removed... Can't see where its used any more.

src/routes/publish.ts
j-maynard: Save for a future refactor.

src/controllers/publish.ts
j-maynard: Fixed

src/controllers/publish.ts
j-maynard: Changed to `continue` and `cancel`

PR #67, opened 2025-01-07
Implement Reference data in the frontend

j-maynard:
Implements reference data in frontend.  Fixes a few bugs and content errors from the previous PR.

### DIFF
--- a/src/controllers/publish.ts
+++ b/src/controllers/publish.ts
@@ -639,17 +639,54 @@ export const fetchDimensionPreview = async (req: Request, res: Response, next: N
                     };
                     res.redirect(req.buildUrl(`/publish/${dataset.id}/lookup/${dimension.id}`, req.language));
                     return;
-                case 'age':
-                case 'ethnicity':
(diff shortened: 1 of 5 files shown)


src/controllers/publish.ts
```diff
@@ -639,17 +639,54 @@ export const fetchDimensionPreview = async (req: Request, res: Response, next: N
...
-                case 'age':
-                case 'ethnicity':
-                case 'geography':
-                case 'religion':
-                case 'sexGender':
+                case 'Geog':
+                case 'Age':
+                case 'Eth':
+                case 'Gen':
+                case 'Rlgn':
```
wheelsandcogs, line 646: These feel like they should be enums eventually.

wheelsandcogs (approved):

src/controllers/publish.ts
j-maynard: I agree this probably needs to be an enum soon... but I think we'll leave it and see where the chips fall on Reference Data.

PR #68, opened 2025-01-10
SW-324: Publisher homepage dataset list

wheelsandcogs:
This updates the publisher homepage to render a table of datasets and their statuses.

Friendly "id" (aka "dataset.code") doesn't exist yet, column is just a placeholder for when it does.

Statuses are best-guess approximation at the moment, once we've done some more work on submit for publishing / make live / dataset updates we can revisit.

Screenshot 2025-01-10 at 13 55 46


j-maynard (approved):
Try again... LGTM 🚀

PR #70, opened 2025-01-10
SW-249: add consumer facing dataset list page

wheelsandcogs:
Adds a new page listing all published datasets that does not require the user to log in.

Currently the links don't go anywhere because we do not yet have a consumer-facing preview page, this will be delivered in an upcoming ticket.

Screenshot 2025-01-14 at 10 10 10


j-maynard (approved):
LGTM 🚀

PR #71, opened 2025-01-13
SW-324: design review fixes for dataset list

wheelsandcogs:
Some colours were missing, and sometimes publishing status was `null` or `undefined` causing broken translation keys.


j-maynard (approved):
LGTM 🚀

PR #74, opened 2025-01-14
Moved dataset routes to /developer

wheelsandcogs:
Moved the dev pages (dataset list and dataset details) under `/developer` and limited access to those with a Marvell email until we have user roles.


j-maynard (approved):
LGTM 🚀

PR #76, opened 2025-01-14
FIX: SW-366 - move cleanup of old fact table to upload

wheelsandcogs:
The previous upload was being deleted before a new one was uploaded, meaning the user could be left in a state of having no uploaded fact table.

This PR moves the removal of the old fact table to happen once we upload a new one.


easternbloc (approved):
LGTM

PR #82, opened 2025-01-20
Make it work on Windows

j-maynard:
Follows the same pattern as on the backend.  The frontend had more minor changes.  Anything requiring node-gyp is gone.  Checked to make sure the code would run on Windows.  Adjusted the redis connection to retry if it fails.


wheelsandcogs:
This has the same issue as the backend PR - all line endings have been saved as Windows format so every file is 100% changed.

wheelsandcogs (approved):

PR #83, opened 2025-01-21
Replace nav links in header

wheelsandcogs:
Fixes a few issues with the frontend nav:

 * "Publish dataset" link removed as it's not in the prototype
 * "List datasets" was moved to "Developer" temp link that only shows for users authed with Marvell emails
 * Guidance link should only display once authed


j-maynard (approved):
LGTM 🚀

PR #85, opened 2025-01-22
Add missing var to example env

wheelsandcogs:
This needs to be in the env file now instead of the package.json scripts.


j-maynard (approved):
Sorry I should have done this... LGTM 🚀

PR #90, opened 2025-01-27
SW-338: Allow user to withdraw publication and revert to tasklist

wheelsandcogs:
This adds a new action that will revert the scheduled dataset publication back to the "Incomplete" or draft status.


j-maynard (approved):
LGTM 🚀

PR #98, opened 2025-01-30
Add support for username only login (TEMP)

wheelsandcogs:
This adds an incredibly basic login form that takes just a username and redirects the user to the backend for the user check. 

**IT IS NOT SECURE and is not intended to be**

This is purely a temporary workaround for testing purposes until we can authenticate WG users via their own EntraID.

To be removed once the testing session is complete.

Screenshot 2025-01-31 at 14 37 27


easternbloc (approved):
LGTM

PR #105, opened 2025-01-31
SW-430: fix broken team selection

wheelsandcogs:
A change was made at some point to `DatasetDTO` that removed the `team_id` prop, which meant the code that pre-populated the selection on the form was not executing.


easternbloc (approved):
LGTM

PR #119, opened 2025-02-08
Improved tasklist performance

wheelsandcogs:
Paired with https://github.com/Marvell-Consulting/statswales-backend/pull/108.

Previously the tasklist was taking quite a while to load, and as it's the core of the publisher workflow, it results in a frustrating experience.

We do not need to load the full dataset with all relations just for the tasklist, so we have a new route with a minimal set of dataset relations loaded (currently metadata & revisions only). We could further improve this by only fetching the latest revision.


j-maynard (approved):
LGTM 🚀

PR #129, opened 2025-02-17
SW-463: fix most recent update & revision sort order on dataset views

wheelsandcogs:
"Most recent update" was showing the wrong date.


j-maynard (approved):
LGTM 🚀

PR #133, opened 2025-02-19
Fix measure preview/review following backend changes

j-maynard:
The backend changes meant the measure preview was having issues especially around null value display.  This fixes many of these issues and makes the table more presentable.  It also filters the table to only show the values for the displayed language rather than both languages.

image

### DIFF
--- a/src/controllers/publish.ts
+++ b/src/controllers/publish.ts
@@ -446,6 +446,10 @@ export const measurePreview = async (req: Request, res: Response, next: NextFunc
         }
 
         const dataPreview = await req.pubapi.getMeasurePreview(res.locals.dataset.id);
+        let langCol = -1;
+        if (dataPreview.headers.find((header) => header.name.toLowerCase().indexOf('lang') > -1)) {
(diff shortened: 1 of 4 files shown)


src/controllers/publish.ts
```diff
@@ -446,18 +446,23 @@ export const measurePreview = async (req: Request, res: Response, next: NextFunc
         }
 
         const dataPreview = await req.pubapi.getMeasurePreview(res.locals.dataset.id);
+        let langCol = -1;
+        if (dataPreview.headers.find((header) => header.name.toLowerCase().indexOf('lang') > -1)) {
+            langCol = dataPreview.headers.findIndex((header) => header.name.toLowerCase().indexOf('lang') > -1);
+        }
```
wheelsandcogs, line 452: You want the index of the header column that contains the string 'lang'?

Is this the same thing (findIndex() returns -1 if not found)?

```suggestion
        const langCol = dataPreview.headers.findIndex((header) => header.name.toLowerCase().includes('lang'))
```

wheelsandcogs (approved):

PR #152, opened 2025-03-10
Implements improved dimension and measure flows

j-maynard:


src/controllers/publish.ts
```diff
@@ -579,13 +587,28 @@ export const uploadLookupTable = async (req: Request, res: Response, next: NextF
...
+                            message: {
+                                key: 'errors.upload.no_csv'
+                            }
+                        }
+                    ],
+                    uploadType: 'lookup',
+                    dimension,
+                    changeLookup: Boolean(dimension.type === 'lookup_table')
+                });
+                return;
```
wheelsandcogs, line 611: The error handling blocks here are very similar, could just throw an error with the necessary message and then to all the repeated stuff in the catch.

src/controllers/publish.ts
```diff
@@ -650,11 +680,19 @@ export const lookupReview = async (req: Request, res: Response, next: NextFuncti
...
                 case 'true':
-                    res.redirect(req.buildUrl(`/publish/${dataset.id}/dimension/${dimension.id}/name`, req.language));
+                    if (revisit)
+                        res.redirect(
+                            req.buildUrl(`/publish/${dataset.id}/dimension-data-chooser/${dimension.id}`, req.language)
+                        );
+                    else
+                        res.redirect(
+                            req.buildUrl(`/publish/${dataset.id}/dimension/${dimension.id}/name`, req.language)
+                        );
```
wheelsandcogs, line 695: Can we use curly brackets on conditions unless it's a 1-liner please

src/views/partials/dimension-preview-table.ejs
```diff
@@ -0,0 +1,42 @@
...
+        <tr>
+            <% row.forEach(function(cell, index) { %>
+                <% if (headers[index].source_type === 'line_number') { %>
+                    <td class="govuk-table__cell line-number"><span class="linespan"><%= cell %></span></td>
+                <% } else { %>
+                    <td class="govuk-table__cell">
+                        <% switch (headers[index].name) {
+                            case 'start_date':
+                        case 'end_date': %>
+                        <%= locals.dateFormat(locals.parseISO(cell.split('T')[0]), 'do MMMM yyyy') %>
```
wheelsandcogs, line 28: If we're only rendering the date part any way, why do we need to split it first?

src/views/partials/measure-preview-table.ejs
```diff
@@ -0,0 +1,43 @@
...
+                    <% if (cell) { %>
+                        <% if (t(`publish.measure_review.column_values.${cell.toString().toLowerCase()}`) === `publish.measure_review.column_values.${cell.toString().toLowerCase()}`) { %>
+                            <%= cell %>
+                        <% } else { %>
+                            <%= t(`publish.measure_review.column_values.${cell.toString().toLowerCase()}`) %>
+                        <% } %>
+                    <% } %>
+                    <% } %>
+                </td>
+            <% }); %>
```
wheelsandcogs, line 39: The nesting is a bit all over the place here and hard to follow - tbh if we need to do this much conditional formatting to the data we should probably do it in the controller before passing it to the view, and leave the view as purely playback.

I realise this might've just moved from elsewhere but a lot of stuff got hand-waved through because it was urgent for MVP.

src/views/publish/measure-revisit.ejs
```diff
@@ -0,0 +1,45 @@
+<%- include("../partials/header", t); %>
+
+<div class="govuk-width-container app-width-container">
+    <!-- <a href="#" class="govuk-back-link">Back</a> -->
```
wheelsandcogs, line 4: We can remove this I think

src/views/publish/upload-lookup.ejs
```diff
@@ -0,0 +1,62 @@
+<%- include("../partials/header", t); %>
```
wheelsandcogs, line 1: Don't need to pass t

src/views/publish/upload-lookup.ejs
```diff
@@ -0,0 +1,62 @@
...
+
+            <div class="govuk-grid-row">
+                <div class="govuk-grid-column-full">
+                    <form method="post" role="continue" enctype="multipart/form-data">
+                        <fieldset class="govuk-fieldset">
+                            <div class="govuk-form-group">
+                                <label class="govuk-label govuk-label--m" for="csv">
+                                    <%= t('publish.upload.form.file.label') %>
+                                </label>
+                                <input class="govuk-file-upload" id="csv" name="csv" type="file" placeholder="Upload Data Files!" accept=".csv,.parquet,.json,.xls,.xlsx,.gz,text/csv,application/vnd.apache.parquet,application/parquet,application/json,application/vnd.ms-excel,application/vnd.openxmlformats-officedocument.spreadsheetml.sheet,application/msexcel,application/gzip">
```
wheelsandcogs, line 36: Accept should probably be a var that comes from a central place of allowed formats.

src/controllers/publish.ts
```diff
@@ -492,19 +495,24 @@ export const measurePreview = async (req: Request, res: Response, next: NextFunc
             langCol = dataPreview.headers.findIndex((header) => header.name.toLowerCase().indexOf('lang') > -1);
         }
 
+        const revisit = Boolean(measure.measure_table && measure.measure_table.length > 0);
+        let page = 'publish/measure-preview';
+        if (req.path.indexOf('change') > -1) page = 'publish/measure-preview';
+        else if (revisit) page = 'publish/measure-revisit';
```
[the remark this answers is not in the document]
j-maynard: No that doesn't work... You need the third case as you need to override the revisit when you go back through the flow.

src/controllers/publish.ts
j-maynard: It is a 1 liner... the issue here is that it hit the line limit so the ESline/Prettier changed the formatting... When your change goes in this will become a 1 line item.

src/views/partials/measure-preview-table.ejs
j-maynard: So the tables are mostly formatting the tables which come from the backend... You could just render the "raw" view of the table but then you loose some of the friendliness of the table.  I don't see there being any gain moving it to the controller level as in effect you'd just end up rendering the whole table in the controller and passing it to the view.  Rather than just building the view based on the raw table.

src/controllers/publish.ts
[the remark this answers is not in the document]
wheelsandcogs: I must be missing something - 

 * L499: default to `publish/measure-preview`
 * L500: basically a no-op because it just sets it to `publish/measure-preview` again if the condition is met
 * L501: set it to 'publish/measure-revisit' if it's a revisit

What's the 3rd case?

src/views/publish/upload-lookup.ejs
j-maynard: Raise a ticket and we can move it out later... This isn't "new" code as such... Its just borrowed from the upload view.

src/views/publish/upload-lookup.ejs
j-maynard: I think most of the header partials pass in t... Again chalk it up as tech debt and we can clear up later

src/controllers/publish.ts
j-maynard: They're similar but they are at points where you must bail out... E.g. no file... or in the case of the catch an error has occurred.  The one in the middle the 400 of course heads you down the "mismatch" path which is a different error page altogether.

Could reduce the code with a function if it its a big thing.

src/views/publish/measure-revisit.ejs
j-maynard: Removed from here and the dimension revisit page.

wheelsandcogs (approved):

PR #158, opened 2025-03-12
Adding management information designation and fixing TOC styling

hdthomps:


wheelsandcogs (approved):

wheelsandcogs:
Selecting the management designation might fail to save until we do some work on the backend.

PR #174, opened 2025-03-18
Update translations and guidance title

j-maynard:


src/i18n/cy.json
```diff
@@ -148,7 +153,6 @@
         "official": "Ystadegau swyddogol",
         "accredited": "Ystadegau swyddogol achrededig",
         "in_development": "Ystadegau swyddogol sy'n cael eu datblygu",
-        "management": "Gwybodaeth reolaethol",
```
wheelsandcogs, line 151: We want to keep this I think.

src/i18n/cy.json
```diff
@@ -627,9 +688,6 @@
             "in_development": {
               "label": "Ystadegau swyddogol sy'n cael eu datblygu"
             },
-            "management": {
-              "label": "Gwybodaeth reolaethol"
-            },
```
wheelsandcogs, line 632: need to keep this.

wheelsandcogs (commented):
Seem to have lost the designation: management still?

PR #176, opened 2025-03-19
SW-557 improving translation import error message

hdthomps:


src/i18n/en.json
```diff
@@ -985,7 +985,7 @@
                     "label": "The file must be in CSV format",
                     "error": {
                         "missing": "Select the populated translation file to upload",
-                        "invalid": "The selected file must match the required format",
+                        "invalid": "The selected file could not be uploaded. Check you have used the exported translation file for this dataset, with the correct entries in the key column.",
```
wheelsandcogs, line 988: Can we say something like "and haven't modified the key column" or words to that effect?

They won't know if they're correct or not, they just need to not touch them :D

wheelsandcogs (approved):

PR #182, opened 2025-03-19
SW-543: run the full stack in docker for local dev

wheelsandcogs:
This allows the full StatsWales stack to be run without needing to manually start or configure any external services (backend, db, file storage, etc) as they will all be started automatically inside Docker.

You could even work entirely offline if you have previously pulled the stack by commenting out `pull_policy: always` in `docker-compose.yml`.

This will also be the basis for running full e2e tests in CI.

Try it: 
`npm run localstack`

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,18 +18,18 @@ Copy the [.env-example](.env-example) file to `.env` and provide the missing val
 
 ## Running the service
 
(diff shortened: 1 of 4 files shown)


j-maynard (approved):
LGTM 🚀

PR #183, opened 2025-03-19
Implement delete a dataset

j-maynard:


AndyDFox (changes requested):
Remove $lightred: #ffaaaa; from src/public/scss/_wg.scss as it's not a GEL colour.

wheelsandcogs (approved):

AndyDFox (approved):

PR #189, opened 2025-03-21
Fixes downloads on conumer and publisher

j-maynard:
As a small added bonus switches from using the id of the dataset/revision to using a slugified version of the title which is easier to find and understand.

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,11 +93,12 @@
     "marked-gfm-heading-id": "^4.1.1",
     "multer": "^1.4.5-lts.1",
     "nanoid": "^3.3.7",
-    "npm": "^11.0.0",
+    "npm": "^11.2.0",
     "passport": "^0.7.0",
     "pino": "^9.4.0",
     "pino-http": "^10.3.0",
     "redis": "^4.7.0",
+    "slugify": "^1.6.6",
     "uuid": "^10.0.0",
     "walk-object": "^4.0.0"
   }
--- a/src/controllers/consumer.ts
+++ b/src/controllers/consumer.ts
@@ -1,6 +1,7 @@
 import { Readable } from 'node:stream';
 
 import { Request, Response, NextFunction } from 'express';
+import slugify from 'slugify';
 
 import { DatasetListItemDTO } from '../dtos/dataset-list-item';
 import { ResultsetWithCount } from '../interfaces/resultset-with-count';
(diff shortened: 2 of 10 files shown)


package.json
```diff
@@ -93,11 +93,12 @@
     "marked-gfm-heading-id": "^4.1.1",
     "multer": "^1.4.5-lts.1",
     "nanoid": "^3.3.7",
-    "npm": "^11.0.0",
+    "npm": "^11.2.0",
```
wheelsandcogs, line 96: Didn't realise we had npm here... should this even be a dep of the project? NPM should be globally installed anywhere we're using it I think?

src/controllers/consumer.ts
```diff
@@ -51,14 +52,21 @@ export const downloadPublishedDataset = async (req: Request, res: Response, next
       throw new NotFoundException('no published revision found');
     }
 
+    let attachmentName: string;
+    if (revision.metadata?.title) {
+      attachmentName = `${slugify(revision.metadata?.title, { lower: true })}-${revision.revision_index > 0 ? `v${revision.revision_index}` : 'draft'}`;
+    } else {
+      attachmentName = `${dataset.id}-${revision.revision_index > 0 ? `v${revision.revision_index}` : 'draft'}`;
```
wheelsandcogs, line 59: A draft version should never be available on the consumer side, and a published dataset will always have a title, so I think we can remove the condition.

src/controllers/publish.ts
```diff
@@ -466,8 +467,15 @@ export const downloadDataset = async (req: Request, res: Response, next: NextFun
       revision = singleLangRevision(revWithMeta, req.language)!;
     }
 
+    let attachmentName: string;
+    if (revision.metadata?.title) {
+      attachmentName = `${slugify(revision.metadata?.title, { lower: true })}-${revision.revision_index > 0 ? `v${revision.revision_index}` : 'draft'}`;
+    } else {
+      attachmentName = `${dataset.id}-${revision.revision_index > 0 ? `v${revision.revision_index}` : 'draft'}`;
+    }
```
wheelsandcogs, line 475: Can we extract this to a helper function? Or move it to `getDownloadHeaders`?

src/views/partials/header.ejs
```diff
@@ -26,6 +25,12 @@
     <link rel="stylesheet" href="/css/accessible-autocomplete.min.css">
     <link rel="stylesheet" href="/css/app.css">
     <meta property="og:image" content="/images/govuk-opengraph-image.png">
+    <meta name="theme-color" content="#323232">
+    <meta name="theme-color" content="#323232" media="(prefers-color-scheme: light)">
+    <meta name="theme-color" content="#323232" media="(prefers-color-scheme: dark)">
+    <meta name="msapplication-navbutton-color" content="#323232">
+    <meta name="apple-mobile-web-app-capable" content="yes">
+    <meta name="apple-mobile-web-app-status-bar-style" content="#323232">
```
wheelsandcogs, line 33: Is this theme styling requested by UX?

wheelsandcogs (approved):

package.json
j-maynard: No idea... it just updated itself though when I did an NPM install

src/controllers/publish.ts
j-maynard: We can probably move it around... I was planning to revisit it when I redo the downloads for the upcoming consumer view tickets.

package.json
wheelsandcogs: Let's remove it, shouldn't be there I think - possibly an accidental `npm i npm` instead of `npm i -g npm`

src/views/partials/header.ejs
j-maynard: It fixes a bug in the styling... It also only affects Safari.  It's not something cover in the GEL at all which feels a bit odd.  However it does stop safaris toolbar turning bright blue when you visit the site.

PR #190, opened 2025-03-21
SW-562: Highlight "Home" and "Guidance" nav items when active page

wheelsandcogs:
Screenshot 2025-03-21 at 12 00 53


j-maynard (approved):
LGTM 🚀

PR #192, opened 2025-03-21
fix weirdness around asset 404s

wheelsandcogs:
There was some strangeness in the logs around missing assets (favicon.svg etc) which were falling through the static router and hitting the full error handler, which was then throwing errors due to missing functions in the 404 template.

This PR moves the static routes before any middleware is applied, and adds a handler for any asset routes that 404 which just responds with a status rather than trying to render the full 404 page.


j-maynard (approved):
LGTM 🚀

PR #196, opened 2025-03-28
Some minor changes to support the backend PR

j-maynard:


src/controllers/publish.ts
```diff
@@ -533,28 +533,25 @@ export const measurePreview = async (req: Request, res: Response, next: NextFunc
         return;
       } catch (err) {
         const error = err as ApiException;
-        logger.debug(`Error is: ${JSON.stringify(error, null, 2)}`);
-        if (error.status === 400) {
+        const viewErr = JSON.parse((error.body as string) || '{}') as ViewErrDTO;
+        logger.debug(`Error is: ${JSON.stringify(viewErr, null, 2)}`);
+        if (viewErr.status === 400) {
```
wheelsandcogs, line 538: We shouldn't ever need to use the body for the http status code - the backend response status code is added as a property of ApiException (`error.status`).

src/exceptions/api.exception.ts
```diff
@@ -2,11 +2,13 @@ export class ApiException extends Error {
   constructor(
     public message: string,
     public status?: number,
-    public body?: string | FormData
+    public body?: string | FormData,
+    public tag?: string
```
wheelsandcogs, line 6: What's "tag" ?

src/views/publish/measure-preview.ejs
```diff
@@ -24,29 +24,29 @@
...
+
+        <div class="govuk-grid-row">
+            <div class="govuk-grid-column-full">
+                <form method="post" role="continue" enctype="multipart/form-data">
+                    <fieldset class="govuk-fieldset">
+                        <div class="govuk-form-group">
+                            <label class="govuk-label govuk-label--m" for="csv">
+                                <%= t('publish.upload.form.file.label') %>
+                            </label>
+                            <input class="govuk-file-upload" id="csv" name="csv" type="file" placeholder="Upload Data Files!" accept=".csv,.parquet,.json,.xls,.xlsx,.gz,text/csv,application/vnd.apache.parquet,application/parquet,application/json,application/vnd.ms-excel,application/vnd.openxmlformats-officedocument.spreadsheetml.sheet,application/msexcel,application/gzip">
```
wheelsandcogs, line 37: Can we move this list of supported formats to a const somewhere please.

src/exceptions/api.exception.ts
j-maynard: Tag is the tag you give to the i18next to replay... While body gives us more information it shouldn't give us the error message to replay to the user as all that is fixed strings.

src/views/publish/measure-preview.ejs
j-maynard: As you request this has now been moved out to an enum.

src/controllers/publish.ts
j-maynard: Switched to using `error.status`

wheelsandcogs (approved):

PR #198, opened 2025-03-28
Updating error messages and dimension guidance

hdthomps:
Welsh version of changes to 'Creating a new dataset' guidance is pending and will be added with separate PR.


j-maynard (approved):
LGTM 🚀

PR #200, opened 2025-04-01
Fix issues around developer view and fact table issues

j-maynard:
Introduces the handling of the bad fact tables.  Fact tables are now validated on the backend as part of the processing of the column sources.  This should prevent a situation where a user makes it as far as creating dimensions and then not being able to proceed with dimension creation and updates.

Also fixes the very broken developer view.  Developer view now loads without issue.  Page needs to be completely migrated to support the new API.  Think the best way to go will be to complete the initial consumer view and then build off the back of that.


src/public/scss/_wg.scss
```diff
@@ -2392,4 +2392,38 @@ $whitepurple: #eed5e3;
...
+    border-bottom: 1px solid #b5b5b5;
+    position: sticky;
+    top: 0;
+  }
+
+  div.alert-warning > details > table > thead > tr > th,
+  div.alert-warning > table > thead > tr > th {
+    background-color: $whitered;
+    border-bottom: 1px solid #b5b5b5;
+  }
```
wheelsandcogs, line 2427:
```
div.alert-warning > details > table > thead > tr > th,
div.alert-warning > table > thead > tr > th {
```

feels incredibly specific, is 

```
div.alert-warning th {
```

too broad?

src/routes/developer.ts
```diff
@@ -20,70 +11,8 @@ developer.use((req: Request, res: Response, next: NextFunction) => {
...
-      res.header('Content-Type', dataTable.mime_type);
-      res.header(`Content-Disposition: attachment; filename="${dataTable.filename}"`);
-      const readable: Readable = Readable.from(fileStream);
-      readable.pipe(res);
-    } catch (_err) {
-      next(new NotFoundException('errors.import_missing'));
-    }
-  }
-);
+developer.get('/:datasetId/import/:factTableId', fetchDataset(Include.All), downloadDataTableFromRevision);
```
wheelsandcogs, line 18: 👍

src/public/scss/_wg.scss
j-maynard: Probably not actually... Although I think following the review I did with Andy and Harry earlier I don't think you should end up in a position where you have a table in an alert box.  Let me remove this.

src/public/scss/_wg.scss
```diff
@@ -2392,4 +2392,38 @@ $whitepurple: #eed5e3;
...
+    padding-left: 20px;
+  }
+
+  td.line-number + td {
+    padding-left: 20px;
+  }
+
+  table {
+    position: relative;
+  }
```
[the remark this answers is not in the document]
wheelsandcogs: It will affect any element inside a table that you want to position outside the flow - and end up with the table as the relative rather than one of the parent elements that you might've intended... I'm not saying we have that currently but it'd be an annoying issue to work around.

Can we at least use a class like `sticky-table` or something?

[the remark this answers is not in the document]
j-maynard: I can make that change

PR #201, opened 2025-04-03
Updated developer view with support for file downloads

j-maynard:
This continues the process of updating the developer view.  Everything in the view now works and this PR adds support for downloading the original files which make up a dataset.  There's options for individual files as well as the option to pull down everything stored in blob storage.

Outstanding tasks left to do is to revisit the Revision and Dimension tabs to make them better and with more information in a better format.


wheelsandcogs (approved):

wheelsandcogs:
I assume we'll be disabling or removing the developer routes for prod?

PR #208, opened 2025-04-08
SW-583: View users in a group

wheelsandcogs:
Updates the group view page to include a list of assigned users.


j-maynard (approved):
LGTM 🤘🏻

PR #210, opened 2025-04-09
SW-644: improvements to user create/update journey

wheelsandcogs:
Improvements requested during design review of the user management journeys.

### DIFF
--- a/src/views/partials/flash-messages.ejs
+++ b/src/views/partials/flash-messages.ejs
@@ -2,9 +2,11 @@
 <div class="govuk-notification-banner govuk-notification-banner--success" role="region" data-module="govuk-notification-banner">
     <div class="govuk-notification-banner__content">
         <% locals.flash?.forEach(function(msg) { %>
-            <p class="govuk-notification-banner__heading">
-                <%= t(msg) %>
-            </p>
+            <% if (typeof msg == 'string') { %>
+                <p class="govuk-notification-banner__heading"><%= t(msg) %></p>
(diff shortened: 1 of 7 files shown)


src/views/partials/flash-messages.ejs
```diff
@@ -2,9 +2,11 @@
...
     <div class="govuk-notification-banner__content">
         <% locals.flash?.forEach(function(msg) { %>
-            <p class="govuk-notification-banner__heading">
-                <%= t(msg) %>
-            </p>
+            <% if (typeof msg == 'string') { %>
+                <p class="govuk-notification-banner__heading"><%= t(msg) %></p>
+            <% } else if (msg != null && typeof msg == 'object') { %>
+                <p class="govuk-notification-banner__heading"><%= t(msg.key, msg.params) %></p>
+            <% } %>
```
wheelsandcogs, line 9: As before, flash messages saving to the session:

```ts
req.session.flash = ['some.translation.key'];
res.session.save();
```
or (new feature - param support)
```ts
req.session.flash = [{ key: 'some.translation.key', params: { user: user.name } }];
req.session.save();
```

and will then be available via:
```ts
const flash = res.locals.flash;
```

in any route that uses the `flashMessages()` middleware, e.g.

```ts
router.get('/my/route', flashMessages, (req, res, next) => {
  // etc...
});
```

j-maynard (approved):
LGTM 🤘🏻

PR #214, opened 2025-04-10
Move to using tsconfig bases

j-maynard:
Some small config changes to move to using tsconfig bases.  This gives us a more recent version of JavaScript to compile to and a number of other languague benefits.  The base specifically targets node22 the version we're using.

Also updated `npm run dev` to use `ts-node-dev` for better source mapping information during development


package.json
```diff
@@ -32,6 +32,7 @@
   "devDependencies": {
     "@eslint/js": "^9.22.0",
     "@playwright/test": "^1.49.1",
+    "@tsconfig/node22": "^22.0.1",
```
wheelsandcogs, line 35: let's remove `nodemon` dep if we've stopped using it 👍

j-maynard: Removed 🤘🏻

wheelsandcogs (approved):

package.json
```diff
@@ -19,7 +19,7 @@
     "check": "npm-run-all prettier:fix lint:fix test build",
     "dev:check": "npm-run-all check dev",
     "predev": "bash -c 'f() { if [ $(command -v podman) ]; then podman compose up -d valkey; else docker compose up -d valkey; fi }; f'",
-    "dev": "nodemon --watch src -e ts,ejs,json,scss --exec npm run start",
+    "dev": "ts-node-dev --respawn --transpile-only src/server.ts | pino-colada",
     "prelocalstack": "bash -c 'f() { if [ $(command -v podman) ]; then podman compose up -d valkey db blobstorage backend; else docker compose up -d valkey db blobstorage backend; fi }; f'",
     "localstack": "trap 'exit 0' SIGINT; nodemon --watch src -e ts,ejs,json,scss --exec npm run start",
```
wheelsandcogs, line 24: We'll need to update the localstack script as well

```suggestion
    "localstack": "trap 'exit 0' SIGINT; ts-node-dev --respawn --transpile-only src/server.ts | pino-colada",
```

wheelsandcogs:
We found issues with moving to ts-node-dev on the frontend (doesn't serve static assets) 

Closing and replacing with https://github.com/Marvell-Consulting/statswales-frontend/pull/234 which is the same thing but without removing nodemon for the time being.

PR #216, opened 2025-04-10
SW-570 - Update to use StatsCymru logo when viewing in Welsh

joefitter:
* Added StatsCymry logo
* Updated  html lang attribute to pull from i18n
* Added scss logo override for cy
* Updated aria label to StatsCymru logo for Welsh


src/i18n/cy.json
```diff
@@ -34,6 +34,7 @@
     "product_name": "StatsCymru F3.0",
     "beta": "Beta",
     "feedback": "Rydych chi'n edrych ar fersiwn newydd o StatsCymru.  Gallwch gael cymorth trwy droi at <a class=\"govuk-link\" href=\"mailto:{{support_email}}\">{{support_email}}</a>.",
+    "logo": "StatsCymru logo",
```
wheelsandcogs, line 37: I think @j-maynard prefers to omit new stuff from the Welsh for when he's collecting stuff for the translators. Best to check.

src/public/scss/_wg.scss
```diff
@@ -43,8 +42,29 @@ $whitepurple: #eed5e3;
...
+  .govuk-tag,
+  .govuk-input,
+  .govuk-label,
+  .govuk-button,
+  .govuk-select,
+  dl,
+  .govuk-accordion__section-heading-text,
+  span,
+  .govuk-inset-text {
+    font-family: Arial, Helvetica, 'Nimbus Sans L', sans-serif;
```
wheelsandcogs, line 67: Your editor prefs are leaking :D

src/views/partials/header.ejs
```diff
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html lang="en" class="govuk-template wg">
+<html lang="<%= i18n.language %>" class="govuk-template wg">
```
wheelsandcogs, line 2: 👍

wheelsandcogs: Please can you make the corresponding change in the consumer side header if it needs it? ta

wheelsandcogs (approved):

src/public/scss/_wg.scss
joefitter: yeah is this a good thing or bad thing? My editor uses prettierconfig and editorconfig, so this is the configured style

joefitter: I'll rebase with the prettier changes once merged in

PR #221, opened 2025-04-11
SW-640: select user group for dataset

wheelsandcogs:
Adds a new page between start -> dataset title screens, that allows the user to select the user group that they belong to that owns the dataset.

If the user is only an editor in a single group, then this page is skipped and that group will be used as the owner for the dataset.

Screenshot 2025-04-11 at 17 44 25


j-maynard (approved):
LGTM 🤘🏻

PR #227, opened 2025-04-15
SW-588: Limit dataset access to users in the assigned group

wheelsandcogs:
User's will now get a 403 Forbidden page if they try and access datasets / revisions / etc owned by groups they are not a member of.


j-maynard (approved):
LGTM 🤘🏻

PR #228, opened 2025-04-15
Implement basic table view

j-maynard:


src/views/publish/dimension-chooser.ejs
```diff
@@ -18,7 +18,7 @@
         <% if (locals?.data) { %>
             <div class="govuk-grid-row">
                 <div class="govuk-grid-column-full with-overflow">
-                    <%- include("../partials/dimension-preview-table.ejs", { dimension: locals.dimension, headers: locals.headers, data: locals.data }) %>
+                    <%- include("../partials/view-table.ejs", { dimension: locals.dimension, headers: locals.headers, data: locals.data }) %>
```
wheelsandcogs, line 21: This file path does not exist in this branch (missing file commit?)

There's a `view-table.ejs` in `views/consumer/partials/` but not in `views/partials/`

src/services/consumer-api.ts
```diff
@@ -81,6 +82,24 @@ export class ConsumerApi {
...
+    logger.debug(`Fetching published view of dataset: ${datasetId}`);
+    let params = '';
+    if (sortBy) {
+      params = `${new URLSearchParams({ page_number: pageNumber.toString(), page_size: pageSize.toString(), sort_by: sortBy }).toString()}`;
+    } else {
+      params = `${new URLSearchParams({ page_number: pageNumber.toString(), page_size: pageSize.toString() }).toString()}`;
+    }
+    return this.fetch({ url: `published/${datasetId}/view?${params}` }).then(
+      (response) => response.json() as unknown as ViewDTO
+    );
```
wheelsandcogs, line 100:
```suggestion
    const searchParams = new URLSearchParams({ page_number: pageNumber.toString(), page_size: pageSize.toString() });
    if (sortBy) searchParams.append(sort_by, sortBy);

    return this.fetch({ url: `published/${datasetId}/view?${searchParams.toString()}` }).then(
      (response) => response.json() as unknown as ViewDTO
    );
```

src/views/publish/dimension-chooser.ejs
j-maynard: Fixed... This might have been a refactoring error

src/views/publish/dimension-revisit.ejs
```diff
@@ -17,7 +17,7 @@
         <% if (locals?.data) { %>
             <div class="govuk-grid-row">
                 <div class="govuk-grid-column-full with-overflow">
-                    <%- include("../partials/dimension-preview-table.ejs", { dimension: locals.dimension, headers: locals.headers, data: locals.data }) %>
+                    <%- include("../partials/view-table.ejs", { dimension: locals.dimension, headers: locals.headers, data: locals.data }) %>
```
wheelsandcogs, line 20: I don't think this path exists on the publish side? Is there a missing file from the commit, or maybe an overly greedy search-replace?

wheelsandcogs (approved):

PR #236, opened 2025-04-22
SW-536: namespace publishing session props using dataset id

wheelsandcogs:
Previously noticed an issue where session props used for controlling flow through the publishing journey leaked between different datasets, which resulted in some weird behaviour.

Solution is to ensure these props are namespaced by the dataset id, so different datasets cannot share session values when the user switches between multiple datasets during the same session.

We could potentially do this for flash and error messages as well, but they're automatically cleared on the very next request already, so are less likely to have similar bleed between datasets.


j-maynard (approved):
LGTM 🤘🏻

PR #248, opened 2025-04-25
Fix broken pagination

j-maynard:
All pages which use the pagination component are broken following the update to the consumer view.  When doing an if on an undefined results in a 500 in ejs.  This updates all uses of pagination to define the variable when calling the partial.


easternbloc (approved):

wheelsandcogs:
I think the pagination issue was resolved without this, so this can be closed without merging right?

PR #254, opened 2025-05-01
Update support link to Slack (Welsh)

wheelsandcogs:
Don't think we need to change the text here, just added the Slack link:
Screenshot 2025-05-01 at 10 46 21


j-maynard (approved):
LGTM 🤘🏻

PR #255, opened 2025-05-01
GEL - SW-298 - Add background to form pages

joefitter:
* Added form-background wrapper to form pages
* Added styling for sticky table headers when in a form-background wrapper


src/views/publish/preview.ejs
```diff
@@ -186,6 +186,6 @@
         letter-spacing: 0.08em;
         line-height: 20px;
         text-transform: uppercase;
-    }
+    } */
```
wheelsandcogs, line 189: If this is all redundant, can we bin it off?

wheelsandcogs (approved):
If we don't need the inline styling on `preview.ejs` please remove it before merging. Otherwise, all good 👍

src/views/publish/preview.ejs
joefitter: good spot, yeah I think it can do, table renders ok

PR #268, opened 2025-05-08
Align developer consumer and publisher views

j-maynard:
This aligns the publisher preview, the consumer view and the developer view to use the same template as their base.  This risks the developer view being a little unstable but improves the publisher and consumer views


src/controllers/developer.ts
```diff
@@ -35,25 +40,29 @@ export const listAllDatasets = async (req: Request, res: Response, next: NextFun
 
 export const displayDatasetPreview = async (req: Request, res: Response) => {
   const dataset = singleLangDataset(res.locals.dataset, req.language);
```
wheelsandcogs, line 42: this is relying on the `fetchDataset()` middlware
```
developer.get('/:datasetId', fetchDataset(Include.All), displayDatasetPreview);
```
which calls an endpoint which includes a user perms check on the backend.

We might want to either add a bypass for users with the developer role, or it might be more explicit to force the developer frontend to use an entirely different endpoint?

src/controllers/developer.ts
```diff
@@ -35,25 +40,29 @@ export const listAllDatasets = async (req: Request, res: Response, next: NextFun
...
+  const revisionId = getLatestRevision(res.locals.dataset)?.id;
+  let pagination: (string | number)[] = [];
+  let errors: ViewError[] | undefined;
+  const revWithMeta = await req.pubapi.getRevision(dataset.id, revisionId!);
```
wheelsandcogs, line 46: Same deal for this request, it'll have a user perms check currently.

src/routes/publish.ts
```diff
@@ -87,8 +87,8 @@ publish.get('/:datasetId/delete', fetchDataset(Include.Meta), deleteDraft);
 publish.post('/:datasetId/delete', fetchDataset(Include.Meta), deleteDraft);
 
 /* Cube Preview */
-publish.get('/:datasetId/cube-preview', fetchDataset(), cubePreview);
-publish.get('/:datasetId/cube-preview/download', fetchDataset(), downloadDataset);
+publish.get('/:datasetId/cube-preview', fetchDataset(Include.All), cubePreview);
+publish.get('/:datasetId/download', fetchDataset(Include.All), downloadDataset);
```
wheelsandcogs, line 91: `Include.All` should not be needed here as it's now part of the `cubePreview` controller:

https://github.com/Marvell-Consulting/statswales-frontend/blob/d4b37f332e52a6a3c4cfca550968b9118d6fbb97/src/controllers/publish.ts#L504

`downloadDataset` doesn't use the dataset at all:
https://github.com/Marvell-Consulting/statswales-frontend/blob/d4b37f332e52a6a3c4cfca550968b9118d6fbb97/src/controllers/publish.ts#L562

I've just spent ages removing these calls as they can add 500ms+ every time they're hit 😄

src/views/partials/header.ejs
```diff
@@ -24,6 +24,10 @@
     <link rel="stylesheet" href="/css/govuk-frontend.min.css">
     <link rel="stylesheet" href="/css/accessible-autocomplete.min.css">
     <link rel="stylesheet" href="/css/app.css">
+    <% if (locals?.isDeveloper) { %>
+        <link rel="stylesheet" href="/css/highlight.css">
+        <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/firacode/6.2.0/fira_code.min.css">
```
wheelsandcogs, line 29: ?

src/controllers/developer.ts
```diff
@@ -96,21 +105,24 @@ export const displayDatasetPreview = async (req: Request, res: Response) => {
...
+    ignoreIllegals: true
+  }).value;
 
-  logger.debug(`Developer view details:\n${JSON.stringify({ ...datasetView, dataset, page, pageSize }, null, 2)}`);
-  res.render('developer/data', {
+  res.status(status);
+  res.render('consumer/view', {
     ...datasetView,
+    datasetMetadata: previewMetadata,
+    ShowDeveloperTab: true,
```
wheelsandcogs, line 117: vars should be camelCase

src/views/consumer/view.ejs
```diff
@@ -1,17 +1,47 @@
-<%- include("partials/header"); %>
-<div class="form-background">
+<% if (locals?.isDeveloper) { %>
+    <%- include("../partials/header"); %>
+<% } else { %>
+    <%- include("partials/header"); %>
+<% } %>
```
wheelsandcogs, line 5: Not a massive fan of having these types of conditions in the consumer templates... 

Previously we had a consumer template that just included partials that were shared between publisher/consumer side, that way there's no chance of this stuff ever being accidentally exposed. Is there a reason we can't still do it that way?

j-maynard: Yes but there's the strong call from UX to use the same view for all data views... Maintaining two near identical views makes very little sense from a development point of view.

src/controllers/developer.ts
j-maynard: Fixed

src/views/consumer/view.ejs
wheelsandcogs: I think that was more about the data /metadata views looking identical. We could just share those and keep the dev-specific stuff in the dev template.

wheelsandcogs: Ok lets merge it for now and revisit.

wheelsandcogs (approved):

PR #276, opened 2025-05-13
SW-571 - GEL - table styles

joefitter:
* Updated font size to 16px for tr
* Removed inline styles
* Removed top-links styles which break horizontal layout - this was causing the side scrolling issue mentioned in the ticket
* Moved top links to header to remove hacky positioning


wheelsandcogs (approved):

wheelsandcogs:
I manually fixed the css map via GitHub UI so it's probably outdated and needs regenerating, but that can wait until the next PR

PR #280, opened 2025-05-16
Bugfix - SW-717 include measure in exported translations

joefitter:
* Updated translations table and i18n to include measure


j-maynard (approved):
LGTM 🤘🏻

PR #288, opened 2025-05-21
change datatable form is now part of the preview page

wheelsandcogs:


j-maynard (approved):
LGTM 🤘🏻 If we're sure this isn't being used any more

PR #289, opened 2025-05-21
SW-675: Set cache-control no-cache header for pub/dev routes

wheelsandcogs:
Users keep complaining about seeing stale statuses in the tasklist, so just force their browser to always refetch.


j-maynard (approved):
LGTM 🤘🏻

PR #291, opened 2025-05-21
add dataset title & publishing date to task-decision

wheelsandcogs:


src/views/publish/task-decision.jsx
```diff
@@ -13,23 +14,50 @@ export default function TaskDecision(props) {
       <div className="govuk-grid-row">
         <div className="govuk-grid-column-two-thirds">
           <form encType="multipart/form-data" method="post">
-            <h1 className="govuk-heading-xl" id="task-decision">
-              <T>publish.task.decision.{props.taskType}.heading</T>
```
joefitter, line 17: is this not working as expected?

wheelsandcogs: reverted back to using `<T>`

joefitter (approved):

PR #309, opened 2025-05-29
Bugfix - template translation errors

joefitter:
* Updated malformatted template strings
* Updated key -> value, key is a reserved prop name in react


j-maynard (approved):
LGTM 🤘🏻

PR #315, opened 2025-06-02
Implement filtering and sorting on all views

j-maynard:
## Summary

This pull request implements filtering and sorting capabilities across all relevant views in the application. Users are now able to filter and sort the data displayed according to various criteria, improving the overall usability and accessibility of the platform.

## Motivation

Previously, users were unable to easily find or organise data within the application views, which could make working with large datasets cumbersome. This enhancement addresses user needs for filtering.

## Implementation Details

- Added generic filtering and sorting logic to the main data views.
- Introduced reusable UI components for filter and sort controls.
- Updated state management to accommodate dynamic filter and sort parameters.
- Ensured that filtering and sorting work in conjunction and update the displayed data in real time.
- Refactored existing view logic to support the new features without breaking legacy functionality.

## Testing

- Manually verified that filtering and sorting options appear and work as expected on all supported views.
- Tested edge cases such as empty filters, reset to default, and sorting with identical values.
- Added/updated unit tests to cover new logic where applicable.

## Notes for Reviewers

- Please check for any performance regressions, especially with large datasets.
- Pay particular attention to the interaction between multiple filters and sorts.
- Let me know if there are any additional filter/sort criteria that should be included.


src/controllers/developer.ts
```diff
@@ -245,3 +245,21 @@ export const downloadAllDatasetFiles = async (req: Request, res: Response, next:
...
+
+export const rebuildCube = async (req: Request, res: Response, next: NextFunction) => {
+  const dataset = singleLangDataset(res.locals.dataset, req.language);
+  const revision = dataset.draft_revision;
+  if (!revision) {
+    logger.error('No draft revision found');
+    next(new NotFoundException('errors.draft_missing'));
+    return;
+  }
+
```
wheelsandcogs, line 257: This route does an `Include.All` but only uses the draft revision id. Hopefully not a regularly used route but just something to keep in mind for other routes.

PR #336, opened 2025-06-23
fix sticky headers for data table

wheelsandcogs:
Removes a style that was stopping the table headers from being sticky.


wheelsandcogs:
@joefitter this style was added here:
https://github.com/Marvell-Consulting/statswales-frontend/commit/f7530ea365ef0c89968dec358e6297ed2b48f457#diff-653f9a85f2b9f1fd2b75a07f71a2b7032150e73a965e0614f9f7f6a2c249bccfR2456 - I assume there was a reason so I've just commented it out for now, but everything I've checked seems to still look ok.

j-maynard (approved):
LGTM 🤘🏻

PR #339, opened 2025-06-26
Feature - SW-827 - batch select filters

joefitter:
* Improved previous logic for JS only elements, now we render hidden with react and unhide and add listeners using client code
* Added select all | none elements to filters


j-maynard (approved):
LGTM 🤘🏻

PR #345, opened 2025-07-04
Feature - SW-472 - sorting

joefitter:
* Updated table component to include sortable headers
* Fixed persistence bugs in buildUrl and pagination components


src/middleware/language-switcher.ts
```diff
@@ -26,7 +27,7 @@ export const localeUrl = (
   // }
 
   const newPath = isEmpty(pathElements) ? '' : `/${pathElements.join('/')}`;
-  const queryString = isEmpty(query) ? '' : `?${new URLSearchParams(query).toString()}`;
+  const queryString = isEmpty(query) ? '' : `?${qs.stringify(query)}`;
```
wheelsandcogs, line 30: what does qs do that URLSearchParams doesn't?

wheelsandcogs (approved):

src/middleware/language-switcher.ts
joefitter: preserves nested structures, else you ended up with things like filters[].a=1,filters[].b=2 etc (it breaks)

PR #346, opened 2025-07-04
if the task was already closed, just redirect back to overview

wheelsandcogs:
There was a bug where refreshing the task decision page after it was already closed left us with broken translations due to the task having the status of withdrawn.

This is not a valid screen so just redirect the user back to the overview page.


j-maynard (approved):
LGTM 🤘🏻

PR #357, opened 2025-07-09
Fix bug in upload and align page sizes

j-maynard:
The page sizes for consumer and publisher have been unaligned for a while.  This small PR aligns the default page sizes to 100 rows.


src/controllers/publish.ts
```diff
@@ -94,6 +94,7 @@ import { SingleLanguageRevision } from '../dtos/single-language/revision';
 import { appConfig } from '../config';
 import { FilterTable } from '../dtos/filter-table';
 import qs from 'qs';
+import { DEFAULT_PAGE_SIZE } from './consumer';
```
wheelsandcogs, line 97: Might be better to stick this in a separate file that both controllers import, so that we're not importing between controllers, but not a blocker.

wheelsandcogs (approved):

PR #366, opened 2025-07-10
Bugfix - SW-846: rows per page select overlaps

joefitter:
* Updated min/max width to 100%, as default min width set by govuk can overflow - this always takes precedence over max-width.


j-maynard (approved):
L... zzzz ... G... zzz... T... 🥱

PR #369, opened 2025-07-11
Accessibility fix - SW-809 page titles missing

joefitter:
* Added page title based on H1 for each page
* Updated page title for update-datatable page on amend journey

### DIFF
--- a/src/views/consumer/list.jsx
+++ b/src/views/consumer/list.jsx
@@ -2,12 +2,16 @@ import React from 'react';
 import Layout from '../components/layouts/Publisher';
 import Pagination from '../components/Pagination';
 
+// TODO: This page is not linked to from anywhere
+// so we should probably remove it.
 export default function ConsumerList(props) {
+  const title = props.t('consumer.list.heading');
   return (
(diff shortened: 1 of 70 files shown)


src/views/consumer/list.jsx
```diff
@@ -2,12 +2,16 @@ import React from 'react';
 import Layout from '../components/layouts/Publisher';
 import Pagination from '../components/Pagination';
 
+// TODO: This page is not linked to from anywhere
+// so we should probably remove it.
```
wheelsandcogs, line 6: It was from before the topics became the main way to find datasets, but it's quite handy to be able to just list published datasets even if we don't link to it, so I would keep it.

src/views/consumer/view.jsx
```diff
@@ -232,9 +232,12 @@ export default function ConsumerView(props) {
     </div>
   );
 
+  const title = props.datasetMetadata.title;
+  console.log({ title });
```
wheelsandcogs, line 236: leftover console.log

wheelsandcogs (approved):

PR #370, opened 2025-07-11
Bugfix - page_size layout broken

joefitter:
* only target sidebar page_size select component when applying full width style


wheelsandcogs:
the css map wasn't committed - might cause conflicts on the next build if it's not.

PR #374, opened 2025-07-15
SW-739: update frequency is now captured as a date input

wheelsandcogs:
Previously we just allowed the user to specify an interval for the next update.

This updates the UI to provide radio buttons to determine the type of update, and a date input to specify the expected update date.

Screenshot 2025-07-15 at 14 29 51


j-maynard (approved):
LGTM 🤘

PR #384, opened 2025-07-22
SW-863: Split the consumer and publisher frontends in to separate servers

wheelsandcogs:
We want to be able to run the consumer and publisher frontends on separate containers in production. This has several benefits:
 - easier to route traffic from subdomains to a specific container
 - a problem (app crash, out of memory, etc) on the publishing side does not take down the public-facing website
 - harder for devs to accidentally expose unpublished data on the consumer side, because the consumer container does not have any publisher URLs routed

This PR splits the code in to 3 directories
 - src/publisher
 - src/consumer
 - src/shared

Generally, code in `src/consumer` should never import anything from `src/publisher`, and (to a lesser extent from a security point of view), vice-versa.

There may still be stuff in `src/shared` that would be better moving to `src/publisher` but this should get us started.


j-maynard (approved):
LGTM 🤘🏻

PR #390, opened 2025-07-24
updated filter all checkbox to have aria-label, added selectable opti…

joefitter:
…ons to number in title


src/shared/views/components/Filters.tsx
```diff
@@ -59,7 +59,11 @@ export const Filters = ({ filters, url, title }: FiltersProps) => {
               <T filtered={filtered} total={total} className={clsx('filtered-label', { 'js-hidden': !filtered })} raw>
                 filters.summary
               </T>
-              <span className={clsx('non-filtered-label', { 'js-hidden': filtered })}>{total}</span>)
+              <span className={clsx('non-filtered-label', { 'js-hidden': filtered })}>
+                {total}
+                <span className="govuk-visually-hidden"> selectable options</span>
```
wheelsandcogs, line 64: Does `selectable options` need to be translated?

PR #399, opened 2025-07-28
hardcode support email instead of using env

wheelsandcogs:
We have the correct support emails now so might as well just hard code them instead of adding extra env vars.


j-maynard (approved):
LGTM 🤘🏻

PR #403, opened 2025-07-29
SW-955: make lang cookie httpOnly

wheelsandcogs:
Per recommendation from the IT Healthcheck - 

Set the language cookie HttpOnly flag so that it cannot be accessed by client-side scripts.


j-maynard (approved):
lgtm 🤘🏻

PR #404, opened 2025-07-29
Updating guidance to reflect current build

hdthomps:


wheelsandcogs (changes requested):
PR looks fine except you've committed some css files that have previously been removed. Please delete your `src/public` dir as it's been moved elsewhere.

wheelsandcogs (approved):

PR #405, opened 2025-07-29
fix react warnings about value prop

wheelsandcogs:
React will whinge about using the prop "value" on inputs:

>Warning: Failed prop type: You provided a `value` prop to a form field without an `onChange` handler. This will render a read-only field. If the field should be mutable use `defaultValue`. Otherwise, set either `onChange` or `readOnly`.

Use "defaultValue" instead and the warning log messages will go away.


j-maynard (approved):
LGTM 🤘🏻

PR #406, opened 2025-07-29
SW-817: show alternate title + [Not yet translated] if not available in current language

wheelsandcogs:
If we do not have a translated title yet, then display the original title along with "Not yet translated" in the appropriate language.

Screenshot 2025-07-29 at 15 06 09

Screenshot 2025-07-29 at 15 06 16


j-maynard (approved):
LGTM... Although dislike it 🤘🏻

PR #409, opened 2025-07-30
SW-696: Cookie banner and settings page

wheelsandcogs:
Adds a banner to the top of the site informing the user about our use of cookies.

Also replaces the cookies page with new text and a form to control settings.

Screenshot 2025-07-30 at 17 46 28


j-maynard (approved):
LGTM generally but we'll need to do things mentioned in Slack and in the comment.

PR #414, opened 2025-07-31
Tweak - SW-903 - updated styles for date fields

joefitter:
Updated date fields to spearate component, import into schedule and year type pages for consistent styling


wheelsandcogs:
There's also https://github.com/Marvell-Consulting/statswales-frontend/blob/main/src/publisher/views/publish/update-frequency.jsx

joefitter:
that wasn't in the ticket :')

PR #423, opened 2025-08-06
SW-953 Disable Geogrpahy reference data type in frontend

j-maynard:
Disables the geogrpahy reference data in the frontend until this can be reworked and improved upon.


src/publisher/views/publish/dimension-chooser.jsx
```diff
@@ -62,10 +62,14 @@ export default function DimensionChooser(props) {
...
-                      label: props.t('publish.dimension_type_chooser.chooser.geography')
-                    },
+                    ...(false
+                      ? [
+                          {
+                            value: 'Geog',
+                            label: props.t('publish.dimension_type_chooser.chooser.geography')
+                          }
+                        ]
+                      : []),
```
wheelsandcogs, line 72: I would just delete this tbh cause it's easy enough to copy from the lines above... who knows if/when this ever comes back

wheelsandcogs (approved):

wheelsandcogs:
Are there any places where we're validating the selection from the form? Because we should remove/disable those as well.

PR #425, opened 2025-08-06
SW-990: replace given/family name with single name

wheelsandcogs:
Backend now just returns a name prop and not a full_name prop.


j-maynard (approved):
LGTM 🤘🏻

PR #430, opened 2025-08-11
SW-989: Render markdown as html in translation previews

wheelsandcogs:
The metadata fields allow the user to provide markdown which is then rendered as HTML in the published dataset.

This PR updates the translation previews to render the markdown as it would appear on the dataset about page.

Screenshot 2025-08-11 at 17 13 11


j-maynard (approved):
LGTM 🤘🏻

PR #432, opened 2025-08-13
SW-857: Dates are now formatted in the dataset table view

wheelsandcogs:
The backend from https://github.com/Marvell-Consulting/statswales-backend/pull/400 onwards will return the correct source type (and date parsing format if relevant) for each column.

We can use this to detect date columns and render them in a nicer way.


j-maynard (approved):
LGTM 🤘🏻

PR #444, opened 2025-08-28
SW-803 Hides period coverage if incomplete or missing

j-maynard:
Bug raised to say period coverage shows as not entered when there's no date reference data used.  This hides that section of the key info if the information isn't present on the dataset.


src/shared/views/components/dataset/KeyInfo.jsx
```diff
@@ -30,6 +31,27 @@ export default function KeyInfo(props) {
...
+              start={dateFormat(props.timePeriod.start, 'MMMM yyyy', props.lang)}
+              end={dateFormat(props.timePeriod.end, 'MMMM yyyy', props.lang)}
+            >
+              dataset_view.key_information.time_period
+            </T>
+          </dd>
+        </div>
+      );
+    }
+    return <div></div>;
```
wheelsandcogs, line 52: Rather than returning an empty div, just return `false` (or `null`, but false is better apparently) and React won't render anything. e.g.

```
if (!props.timePeriod.start || !props.timePeriod.end) {
  return false;
}

return <actual-component>;
```

wheelsandcogs (approved):

PR #453, opened 2025-09-16
SW-1024: fix download metadata for non-published datasets

wheelsandcogs:
The download metadata button was broken when previewing datasets because it was attempting to use a published dataset route when the dataset wasn't published yet.

Recommend to turn off whitespace changes when reviewing, most of the view changes are indentation.


j-maynard (approved):
LGTM 🤘🏻

PR #469, opened 2025-09-26
let i18next decide the cookie domain

wheelsandcogs:
Lang cookie is not being set on Welsh language domain... try removing the forced cookie domain option and let the middleware determine the cookie domain itself.


j-maynard (approved):
LGTM 🤘🏻

PR #476, opened 2025-09-30
SW-1049: add dataset pagination to topic pages

wheelsandcogs:
Adds missing pagination to the dataset lists on topic pages in the consumer view.


j-maynard (approved):
LGTM 🤘🏻

PR #488, opened 2025-10-07
Liveness should only prove the process/event loop is healthy

easternbloc:
no external deps. Keep /ready for dependency checks.


j-maynard (approved):
LGTM 🤘🏻

PR #496, opened 2025-10-27
SW-773: add sorting options to the topic dataset list

wheelsandcogs:
Users can now sort the dataset list by title (asc or desc), first published, or most recently updated. The form will submit automatically on select change.

Screenshot 2025-10-27 at 11 26 56


j-maynard (approved):
LGTM 🤘🏻

PR #497, opened 2025-10-27
SW-1064: Fix issues with nested filters

wheelsandcogs:
Filters with more than 4-5 levels of nesting were not being handled correctly (filter not fully correctly applied, filter not marked as selected after form submission).

This hopefully resolves those issues and should work for any level of nesting.


j-maynard (approved):
LGTM 🤘🏻

PR #503, opened 2025-11-06
SW-976: update frequency should only enforce a year

wheelsandcogs:
Previously we required the publisher to specify at least a month and a year, but this is too restrictive for some datasets. Now we only require a year.

Fixed a bit of weirdness around date validation where 30th Feb was passing as a valid date.

e2e test cases added.


j-maynard (approved):
LGTM 🤘🏻

PR #504, opened 2025-11-12
Add support for long cube builds

j-maynard:
End points which regenerate the cube now return always return a build ID which we can query with the frontend.  If a build takes more than 10 seconds (up for discussion) a page is rendered to say the cube is building.  If it takes less than 10 seconds it automatically redirects to the originally intended page.


src/shared/public/assets/js/long-build.js
```diff
@@ -0,0 +1,63 @@
...
+    sleep(2000);
+  }
+}
+
+intervalId = setInterval(checkBuildStatus, 1000);
+document.getElementById('action-button').style.visibility = 'hidden';
+document.getElementById('build-status-message').innerHTML = translations.message.building.replace(
+  '%time%',
+  countDownTime
+);
```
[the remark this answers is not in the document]
j-maynard: See above... We need rich html support to support Arias on the countdown span.  Additionally there is no direct user input here as the strings originate in our server side translation layer.

src/publisher/services/publisher-api.ts
```diff
@@ -437,14 +438,17 @@ export class PublisherApi {
     }).then((response) => response.json() as unknown as FilterTable[]);
   }
 
-  public async assignSources(datasetId: string, sourceTypeAssignment: SourceAssignmentDTO[]): Promise<DatasetDTO> {
+  public async assignSources(
+    datasetId: string,
+    sourceTypeAssignment: SourceAssignmentDTO[]
+  ): Promise<{ dataset: DatasetDTO; build_id: string }> {
```
wheelsandcogs: Can we have a new interface for this return type please - maybe `DatasetBuildDTO`?

src/@types/express-session/index.d.ts
```diff
@@ -17,5 +17,7 @@ declare module 'express-session' {
     dataset: {
       [id: string]: DatasetProps;
     };
+    buildNextAction: string | undefined;
+    buildPreviousAction: string | undefined;
```
wheelsandcogs: These should be namespaced under the `dataset` key above by placing them in `DatasetProps`, otherwise different datasets can interfere (especially in tests running in parallel for the same user).

src/publisher/controllers/publish.ts
```diff
@@ -715,8 +720,13 @@ export const measurePreview = async (req: Request, res: Response, next: NextFunc
         const fileName = req.file.originalname;
         req.file.mimetype = fileMimeTypeHandler(req.file.mimetype, req.file.originalname);
         const fileData = new Blob([req.file.buffer as BlobPart], { type: req.file.mimetype });
-        await req.pubapi.uploadMeasureLookup(dataset.id, fileData, fileName);
-        res.redirect(req.buildUrl(`/publish/${dataset.id}/measure/review`, req.language));
+        const viewDTO = await req.pubapi.uploadMeasureLookup(dataset.id, fileData, fileName);
+        req.session.buildPreviousAction = req.originalUrl;
+        req.session.buildNextAction = req.buildUrl(`/publish/${dataset.id}/measure/review`, req.language);
```
wheelsandcogs: As above, needs namespacing.

src/publisher/views/publish/long-build.jsx
```diff
@@ -0,0 +1,109 @@
...
+
+export default function LongBuild(props) {
+  const spinner = {
+    display: 'inline-block',
+    height: '50px',
+    width: '50px',
+    verticalAlign: 'middle',
+    marginLeft: '50px'
+  };
+  const showSpinner = () => {
```
wheelsandcogs: When using react components, it's better to make them standalone rather than nesting them inside other components and relying on parent props. It's cleaner and makes it easier to share them with other templates in future. It's also better for performance reasons when the templates are rendered client-side (not really an issue here) so it's a good habit in general.

e.g. move this outside of the parent component `LongBuild`:

```tsx
function ShowSpinner({ buildLogEntry }) {
  ...
}
```

and call it like 

```tsx
   <ShowSpinner buildLogEntry={buildLogEntry} />
```

or just pass all the props if you're using lots of stuff.

```tsx
  <ShowSpinner props={...props} />
```

src/publisher/views/publish/long-build.jsx
```diff
@@ -0,0 +1,109 @@
...
+  };
+  const actionButton = () => {
+    switch (props.buildLogEntry.status) {
+      case CubeBuildStatus.Completed:
+        return (
+          <>
+            <a href={props.nextAction} className="govuk-button" id="action-button">
+              {props.t('buttons.continue')}
+            </a>
+          </>
```
wheelsandcogs: You only need react fragments `<>` if there's more than 1 top level element being returned, e.g. 

```tsx
  return (
    <div>something</div>
  );
```

doesn't required them, but 

```tsx
  return (
    <>
      <div>something</div>
      <div>something else</div>
    </>
  );
```
does.

src/shared/public/assets/js/long-build.js
```diff
@@ -0,0 +1,78 @@
...
+    typeof url === 'string' &&
+    (url.startsWith('/') || url.startsWith('./') || url.startsWith('http://') || url.startsWith('https://'))
+  ) {
+    return url;
+  }
+  // Anything else is potentially dangerous
+  return '#';
+}
+
+async function fetchBuildStatusAndUpdate() {
```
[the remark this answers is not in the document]
j-maynard: Sadly not... I considered this approach but because you can't use meta-refresh for accessibility reasons.  So the alternative was client side javascript to update the pages contents and handle checking the status.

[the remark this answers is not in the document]
j-maynard: You want to use an Ajax request to fetch the new content from the server side... Could do that.. Currently we just return the build object but guess there's nothing stopping us returning a fragment instead

src/shared/views/components/Spinner.tsx
```diff
@@ -1,5 +1,5 @@
-import React from 'react';
+import React, { CSSProperties } from 'react';
 
-export const Spinner = () => {
-  return <div className="loader" aria-live="polite" role="status"></div>;
+export const Spinner = (style?: CSSProperties) => {
+  return <div className="loader" aria-live="polite" role="status" style={style}></div>;
```
wheelsandcogs: Pass in an optional css class, not inline styles.

See https://github.com/Marvell-Consulting/statswales-frontend/blob/main/src/shared/views/components/Select.tsx#L34

src/publisher/views/publish/long-build-fragment.jsx
```diff
@@ -0,0 +1,90 @@
...
+    }
+  };
+
+  const showSpinner = () => {
+    switch (props.buildLogEntry.status) {
+      case CubeBuildStatus.Completed:
+      case CubeBuildStatus.Failed:
+        return <></>;
+      default:
+        return <div className="loader" aria-live="polite" role="status" style={spinnerStyle}></div>;
```
wheelsandcogs:
```jsx
  return <Spinner className="loader--centred" />
```

PR #507, opened 2025-11-12
SW-1065: sanitise dataset titles when used as filename

wheelsandcogs:
Non-ascii characters in filenames can cause issues with content-disposition headers, so use a simplified version of the title instead.


j-maynard (approved):
LGTM 🤘🏻

PR #509, opened 2025-11-14
SW-1078: avoid using dataset.revisions where possible

wheelsandcogs:
This is a partner PR to https://github.com/Marvell-Consulting/statswales-backend/pull/506.

It's very rare that we actually need to load the full list of revisions. Most of the time there are more efficient ways to find the latest / end / draft revision.

This should hopefully speed things up and help with memory consumption.

### DIFF
--- a/tests/publisher-api.test.ts
+++ b/tests/publisher-api.test.ts
@@ -105,25 +105,6 @@ describe('PublisherApi', () => {
     });
   });
 
-  describe('confirmDataTable', () => {
-    it('should return a DataTableDTO', async () => {
-      const datasetId = randomUUID();
-      const revisionId = randomUUID();
-      const importId = randomUUID();
-      const dataTable = { dataset_id: datasetId, revision_id: revisionId, import_id: importId };
-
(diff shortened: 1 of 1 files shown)


src/publisher/services/publisher-api.ts
```diff
@@ -333,15 +333,6 @@ export class PublisherApi {
...
   }
 
-  public async confirmDataTable(datasetId: string, revisionId: string): Promise<DataTableDto> {
-    logger.debug(`Confirming data table import for revision: ${revisionId}...`);
-
-    return this.fetch({
-      url: `dataset/${datasetId}/revision/by-id/${revisionId}/data-table/confirm`,
-      method: HttpMethod.Patch
-    }).then((response) => response.json() as unknown as DataTableDto);
-  }
```
wheelsandcogs: This was a no-op on the backend, it doesn't actually do anything other than return an object that we then don't even assign to a var on the frontend. I assume it was previously doing something more than this and as code was removed it became redundant.

j-maynard (approved):
LGTM 🤘🏻

PR #512, opened 2025-11-25
SW-600: collect and display reason meta for updates

wheelsandcogs:
Collects a reason for each dataset update as part of the metadata.

The reason is played back on a new "Dataset History" tab on the published dataset.

Screenshot 2025-11-25 at 14 39 25

### DIFF
--- a/tests-e2e/publish/unpublish-dataset.spec.ts
+++ b/tests-e2e/publish/unpublish-dataset.spec.ts
@@ -5,7 +5,7 @@ import { nanoid } from 'nanoid';
 
 import { users } from '../fixtures/logins';
 import { config } from '../../src/shared/config';
-import { publishMinimalDataset } from './helpers/publishing-steps';
+import { completeUpdateReason, publishMinimalDataset } from './helpers/publishing-steps';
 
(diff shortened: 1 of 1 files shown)


src/publisher/services/publisher-api.ts
```diff
@@ -825,4 +825,11 @@ export class PublisherApi {
     logger.debug(`Fetching build log entry with id ${buildId}...`);
     return this.fetch({ url: `build/${buildId}` }).then((response) => response.json() as unknown as BuildLogEntry);
   }
+
+  public async getPublicationHistory(datasetId: string): Promise<RevisionDTO[]> {
+    logger.debug(`Fetching publication history for dataset: ${datasetId}`);
+    return this.fetch({ url: `v1/${datasetId}/history` }).then(
+      (response) => response.json() as unknown as RevisionDTO[]
+    );
+  }
```
wheelsandcogs: This is a copy of the function in consumer api - seemed pointless replicating the endpoint on the backend publisher routes, but we don't include the consumer api in the publisher FE services. Maybe we should instead of duplicating this...

PR #515, opened 2025-11-26
Translate update notes

wheelsandcogs:


j-maynard (approved):
LGTM 🤘🏻 
Though might still want @hdthomps to check it as its content.

PR #540, opened 2026-01-28
Add comprehensive dataset filtering tests to consumer e2e suite

easternbloc:
Extend consumer test coverage from 32 to 53 tests with new test suites:
  - Filter sidebar UI (apply button, page size selector, filter groups)
  - Page size changes and persistence
  - Filter application and redirect to /filtered URL
  - Filter controls (select all, none, not filtered checkbox)
  - Pagination navigation and URL parameters
  - Table sorting with URL parameters
  - Download options with filter state preservation

  Also update playwright config to conditionally depend on publish tests
  only in CI (process.env.CI), allowing local testing without running
  the full publish test suite.


tests-e2e/consumer/dataset-filtering.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+
+  test.describe('Applying Filters', () => {
+    test('Applying filters redirects to filtered URL', async ({ page }) => {
+      await page.goto('/en-GB/all');
+      const firstDataset = page.locator('.index-list__item a').first();
+      await firstDataset.click();
+      // Apply filters (even without changing anything, it should work)
+      await page.getByRole('button', { name: 'Apply' }).click();
+      // URL should change to include /filtered/ path
+      await expect(page).toHaveURL(/\/filtered\//);
```
wheelsandcogs: We probably want to check the content on the page is actually the data table and not an error message.

wheelsandcogs (approved):

PR #654, opened 2026-05-20
SW-1276: add a hint explaining that auto redirect also hides the dataset from search / topics

wheelsandcogs:
https://github.com/Marvell-Consulting/statswales-backend/pull/680 makes it so that datasets that are archived, have a replacement dataset set and auto-redirect enabled are now filtered from the search results and topic listings.

This PR updates the content on the archival form to inform the publisher that this will happen if they tick the auto-redirect checkbox.


j-maynard (approved):
LGTM 🤘🏻

PR #656, opened 2026-05-20
SW-1278: assert translation export item stays complete after import

wheelsandcogs:
## Summary

`completeTranslations()` in the publish-dataset Playwright helper previously only asserted that **Import translations** showed Completed after a successful import. The reported bug is the reverse — **Export text fields for translation** flipping back to Incomplete immediately after import — which means the existing assertion would not have caught it.

Added a second assertion right after the existing import check, so a regression of the export status would now fail the publish-dataset spec.

[SW-1278]: https://marvellconsulting.atlassian.net/browse/SW-1278?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


j-maynard (approved):
LGTM 🤘🏻

PR #659, opened 2026-05-26
Remove unused jest-junit reporter

wheelsandcogs:
## Summary

- Removed the `jest-junit` reporter from `jest.config.ts`. It was emitting `coverage/test-report/junit-report.xml` on every test run, but no CI workflow in `.github/workflows/` consumed the file.
- Uninstalled `jest-junit` and `jest-junit-reporter` (the latter being an unmaintained, unreferenced duplicate last published in 2016).

Jest's default reporter is still active (it's the implicit fallback when `reporters` is unset), and code coverage is unaffected — `collectCoverage` and the `cobertura`/`lcov`/`html`/`text` coverage reporters remain configured, as do the coverage thresholds.

## Test plan

- [x] `npm run check` passes (192 tests, build succeeds)
- [x] Confirmed no references to `jest-junit` remain in the repo
- [x] Confirmed default Jest test output still renders in the test run

### DIFF
--- a/jest.config.ts
+++ b/jest.config.ts
@@ -9,7 +9,6 @@ const config: Config = {
     '^(\\.{1,2}/.*)\\.js$': '$1'
   },
   verbose: true,
-  reporters: ['default', ['jest-junit', { outputDirectory: 'coverage/test-report', outputName: 'junit-report.xml' }]],
   testEnvironment: 'node',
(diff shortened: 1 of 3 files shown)


j-maynard (approved):
LGTM 🤘🏻

PR #667, opened 2026-06-03
SW-1059: Show both Welsh and English published URLs in the overview

wheelsandcogs:
### Problem

The dataset overview page shows a future "Once published, the dataset will be available at: <url>" line for unpublished datasets (`datasetStatus === 'new'`), but it only rendered a single URL built from the current page language. A publisher viewing the page in English saw only the English URL and had to switch the whole page to Welsh to grab the Welsh one.

### Change

- The overview page now builds both an English (`/en-GB/.../start`) and Welsh (`/cy-GB/.../start`) public URL and renders an intro line plus two labelled lines (English / Welsh), shown regardless of the current page language.
- The Welsh URL uses `consumer.welshUrl || consumer.url` as its base, matching the canonical-URL pattern already used in the consumer `Layout.tsx`.

### Bonus fix

`cy.json` was missing the `publish.overview.group` and `publish.overview.live_url` keys, so Welsh overview pages were falling back to the English strings for those two lines. Added the missing Welsh translations.

### i18n note

The new Welsh strings reuse existing terms (`Grŵp`, `Saesneg`, `Cymraeg`) but should be confirmed by a Welsh speaker before merge.

### Tests

Added `tests/publish-overview.test.ts` — asserts both URLs render in both `/en-GB` and `/cy-GB`, and that the Welsh labels resolve (not English fallback).

`npm run check` passes (prettier, lint, 269 tests, build).

### DIFF
--- a/src/publisher/views/publish/overview/overview.jsx
+++ b/src/publisher/views/publish/overview/overview.jsx
@@ -16,7 +16,8 @@ export default function Overview(props) {
   const openUnpublishTask = props.openTasks.find((task) => task.action === 'unpublish');
   const openArchiveTask = props.openTasks.find((task) => task.action === 'archive');
   const openUnarchiveTask = props.openTasks.find((task) => task.action === 'unarchive');
-  const eventualPublicUrl = `${config.frontend.consumer.url}/${props.i18n.language}/${props.dataset.id}/start`;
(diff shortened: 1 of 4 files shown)


j-maynard (approved):
LGTM 🤘🏻

PR #675, opened 2026-07-17
SW-935 Fix consumer home and topic screen styling [GEL QA]

j-maynard:
Use padding instead of margin for the hero band's top/bottom spacing so it can't collapse through unstyled ancestor divs and escape the blue background - this was leaving the required 20px+ gap outside the hero rather than around the text on screens below 640px.

Change the topic-subhead label from an h2 to a span so it no longer introduces a heading level before the page's h1, and tightens up against the title to match other eyebrow-label usage in the app.

### DIFF
--- a/src/consumer/views/topic-list.jsx
+++ b/src/consumer/views/topic-list.jsx
@@ -140,9 +140,9 @@ export default function TopicList(props) {
       {!props.selectedTopic && (
         <Hero>
           <div className="govuk-width-container">
-            <div className="govuk-grid-row govuk-!-margin-bottom-6">
+            <div className="govuk-grid-row govuk-!-padding-bottom-6">
               <div className="govuk-grid-column-two-thirds">
-                <h1 className="govuk-heading-xl govuk-!-margin-top-6">{title}</h1>
+                <h1 className="govuk-heading-xl govuk-!-padding-top-6">{title}</h1>
               </div>
             </div>
           </div>
@@ -167,9 +167,9 @@ export default function TopicList(props) {
           {props.selectedTopic && (
             <>
               <ParentTopicBreadcrumbs {...props} />
-              <h2 className="topic-subhead">
+              <span className="topic-subhead">
                 <T>{props.parentTopics.length > 0 ? 'consumer.topic_list.sub_topic' : 'consumer.topic_list.topic'}</T>
-              </h2>
+              </span>
               <h1 className="govuk-heading-xl">{title}</h1>
             </>
           )}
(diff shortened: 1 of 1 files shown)


src/consumer/views/topic-list.jsx
```diff
@@ -167,9 +167,9 @@ export default function TopicList(props) {
           {props.selectedTopic && (
             <>
               <ParentTopicBreadcrumbs {...props} />
-              <h2 className="topic-subhead">
+              <span className="topic-subhead">
                 <T>{props.parentTopics.length > 0 ? 'consumer.topic_list.sub_topic' : 'consumer.topic_list.topic'}</T>
-              </h2>
+              </span>
```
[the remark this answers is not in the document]
j-maynard: Fixed in commit [f46f935](https://github.com/Marvell-Consulting/statswales-frontend/pull/675/commits/f46f9351cfe46768ee136f14d1a9b597baea417c)

src/consumer/views/topic-list.jsx
```diff
@@ -167,9 +167,9 @@ export default function TopicList(props) {
           {props.selectedTopic && (
             <>
               <ParentTopicBreadcrumbs {...props} />
-              <h2 className="topic-subhead">
-                <T>{props.parentTopics.length > 0 ? 'consumer.topic_list.sub_topic' : 'consumer.topic_list.topic'}</T>
-              </h2>
+              <T className="topic-subhead">
+                {props.parentTopics.length > 0 ? 'consumer.topic_list.sub_topic' : 'consumer.topic_list.topic'}
```
easternbloc, line 171: Is the intention for this to now live outside of a heading? It's just a floating span now I think.

easternbloc: Is the topic-subhead no longer a heading?

easternbloc (changes requested):
Is the intention here to not have the topic subhead as part of the heading anymore? as this is now a span I think? Is that intended behaviour?

src/consumer/views/topic-list.jsx
j-maynard: Semantically it shouldn't be a heading... In HTML you should only have 1 H1 on a page and it should always be the first heading on the page.  So starting with an H2 and then having an H1 is a big no where semantic HTML is concerned.  So we preserve the style and correct the structure of the document.

easternbloc: My point is more around the name of the class...

In govuk this would be a `govuk-caption`, does anything exist in the GEL for this?

j-maynard: A new tech debt Jira ticket has been created as SW-1328 to address the lack of naming conventions in the current SCSS files.  Suggests pre-fixing styles with `gel-` and applying the govuk naming conventions to the SCSS files.

PR #681, opened 2026-08-06
Fix broken PR pipelines

j-maynard:
The frontend PR pipelines broke due to the removal of npm and associated files from the backend docker image.  This removes the dependance on npm from the docker-compose file used for our pipelines.


easternbloc:
The surpressed comment from copilot is valid

easternbloc (changes requested):
Just fix up the copilot comment and then it looks good

easternbloc (approved):

